PR #15, opened 2020-06-17
Added extra snapshot tests for validator

conner-jowitt:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ test:
 	&& export API_TEST_ENV_FILE_PATH=$(or $(API_TEST_ENV_FILE_PATH),tests/e2e/environments/local.postman_environment.json) \
 	&& export API_TEST_URL=$(or $(API_TEST_URL),localhost:9000) \
 	&& npm run test
+	cd sandbox && npm t
 
 release: build-spec build-sandbox build-proxy
 	mkdir -p dist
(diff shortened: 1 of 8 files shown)


sandbox/tests/validators/request-validator.spec.js
```diff
@@ -0,0 +1,32 @@
+const validator = require("../../validators/request-validator")
+const validPrescriptionBundle = require("../resources/valid-bundle.json")
+const validPrescriptionBundleWithSignature = require("../resources/valid-bundle-with-signature.json")
+const {Success} = require('monet');
+
+test('it should return an array of validation error objects when given an empty json object', () => {
```
booshibooshi, line 6: I like the naming convention you've gone with for new tests, can this test name be updated to match?

Makefile
```diff
@@ -91,6 +91,7 @@ test:
 	&& export API_TEST_ENV_FILE_PATH=$(or $(API_TEST_ENV_FILE_PATH),tests/e2e/environments/local.postman_environment.json) \
 	&& export API_TEST_URL=$(or $(API_TEST_URL),localhost:9000) \
 	&& npm run test
+	cd sandbox && npm t
```
booshibooshi, line 94: This change introduces a dependency on build-spec and build-sandbox make targets. So if you were to run `make clean` and then `make test` the tests would fail. 

I think we can add these deps to the test makefile target i.e. `test: build-spec build-sandbox`
and remove them from the release target `release: build-proxy`. Then you can run test independently.  

The test and release targets get called successively in azure-pipelines so the CI should continue to function as it does now

PR #16, opened 2020-06-18
Aea 228 update azure pipeline

conner-jowitt:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


Makefile
```diff
@@ -86,7 +86,7 @@ lint:
 validate: build-spec build-sandbox
 	java -jar bin/org.hl7.fhir.validator.jar build/examples/**/*application_fhir+json*.json -version 4.0.1 -tx n/a | tee /tmp/validation.txt
 
-test: build-spec build-sandbox
```
ri-ch, line 89: Are going to check whether the sandbox is already up before we try and run the tests or are we happy to let the tests fail on the first attempt if its not up yet? (either is good, just checking we're on the same page)

conner-jowitt: I think so yes. I added a step in to check if the sandbox was running and to fail if it wasn't but given that the tests will fail if it isn't up anyway it seemed unnecessary. I can add it in if we think we need it though?

PR #39, opened 2020-07-03
Aea 304 update course of therapy type mappings

booshibooshi:
## Summary
* Routine Change
* :sparkles: New Feature

## Reviews Required
* [x] Dev
* [ ] Test
* [ ] Tech Author
* [ ] Product Owner

## Review Checklist


coordinator/src/services/fhir-resources.ts
```diff
@@ -29,6 +29,7 @@ export class MedicationRequest extends Resource {
     authoredOn?: string
     requester?: Reference<PractitionerRole>
     groupIdentifier?: MedicationRequestGroupIdentifier
+    courseOfTherapyType?: TherapyType
```
mfjarvis, line 32: This is a CodeableConcept (we have a class for it which we can use)

coordinator/src/services/translation-service.ts
```diff
@@ -258,6 +256,30 @@ function convertProduct(medicationCodeableConcept: fhir.CodeableConcept) {
     return new prescriptions.Product(manufacturedProduct);
 }
 
+export function convertCourseOfTherapyType(medicationRequests: Array<fhir.MedicationRequest>): prescriptions.PrescriptionTreatmentType {
+    const courseOfTherapyTypeCodes = new Set(
```
mfjarvis, line 260: Could adapt getCodeableConceptCodingForSystem to do this (takes array of CodeableConcepts and a code system, currently returns the single result - could instead make a Set to return the single **unique** result)

mfjarvis: Or could just get the value from the first MedicationRequest in the bundle as we do for various other fields (perhaps not a good idea)

models/schemas/CourseOfTherapyType.yaml
```diff
@@ -0,0 +1,20 @@
+type: object
```
mfjarvis, line 1: Rename to have CodeableConcept at the start of the filename?

mfjarvis: Including the misspelling - it's in the spec

specification/electronic-prescription-service-api.yaml
```diff
@@ -218,365 +217,373 @@ paths:
...
+                type: string
+                example:
+                  $ref: components/examples/ConvertSuccessResponse.xml
         # Emulate the response example rendered in internal-portal whilst this response model is being iterated on
         # TODO: Add yaml representation of the model to models/responses with conversion to xml to models/dist/responses
         # for use in tests
-        '200':
-          description: |
-            Successfully converted prescription.
+        # '200':
```
mfjarvis, line 234: Comment can go now

PR #112, opened 2020-08-07
updated demographics tests

conner-jowitt:


coordinator/tests/services/translation/demographics.spec.ts
```diff
@@ -112,35 +78,35 @@ describe("convertAddress should return correct addresses", () => {
...
 })
 
 describe("convertGender should return correct gender", () => {
-  test("valid fhirGender returns correct hl7 gender", () => {
-    expect(demographics.convertGender("male")).toEqual(codes.SexCode.MALE)
-    expect(demographics.convertGender("female")).toEqual(codes.SexCode.FEMALE)
-    expect(demographics.convertGender("other")).toEqual(codes.SexCode.INDETERMINATE)
-    expect(demographics.convertGender("unknown")).toEqual(codes.SexCode.UNKNOWN)
-  })
+  const cases = [["male", codes.SexCode.MALE],
```
ri-ch, line 101: Minor point, but this array should be formatted the same was as the others (with first entry on a new line)

coordinator/tests/services/translation/demographics.spec.ts
```diff
@@ -51,22 +38,14 @@ describe("convertName fills correct fields only", () => {
...
-    const result = demographics.convertName(fhirName)
-    expect(result._attributes).toEqual({use: core.NameUse.USUAL})
-  })
-
-  test("official should return USUAL", () => {
-    const fhirName = {"use": "official"}
-    const result = demographics.convertName(fhirName)
-    expect(result._attributes).toEqual({use: core.NameUse.USUAL})
-  })
+  const cases = [["usual", core.NameUse.USUAL],
```
ri-ch, line 41: Just because i think it might look neater, could we format these arrays like this:
```
const cases = [
  ["one", 1],
  ["two", 2]
]
```

PR #120, opened 2020-08-12
AEA-458: Handle direct response to a Send request

mfjarvis:


coordinator/src/services/spine-communication.ts
```diff
@@ -65,26 +64,10 @@ export class RequestHandler {
...
-          statusCode: result.status,
-          pollingUrl: pollingUrl
-        }
-      }
-      default: {
-        console.log(`Got the following response from spine:\n${result.data}`)
-        throw Error(`Unsupported status, expected 202, got ${result.status}`)
-      }
-      }
+      return RequestHandler.handlePollableOrImmediateResponse(result)
```
ri-ch, line 67: I like the refactoring, the only problem with this approach is that a 200 response from Spine does actually represent an error, as far as the client is concerned(so it should probably be a 400). Maybe this is something we can tidy up when we have some means of processing spine responses to work out whether they represent a success or not?

mfjarvis: Yeah, this is an existing problem. We're currently returning a 200 for errors, even from the polling endpoint. The only way to know whether we should return a 200 or a 400 is to parse the application acknowledgement from Spine. This should be fixed by AEA-368.

ri-ch: Fair enough. We should probably update the spec to reflect the (potential) change in behaviour

PR #121, opened 2020-08-12
Aea 492 rename endpoints

conner-jowitt:


specification/electronic-prescription-service-api.yaml
```diff
@@ -133,7 +133,7 @@ tags:
   - name: dispensing
   - name: prescribing
 paths:
-  /Prepare:
```
mfjarvis, line 136: Github won't let me put a comment there but lines 87 and 125-127 need changing too

conner-jowitt: Removed line 87, updated the other lines

PR #122, opened 2020-08-13
Add timezone tests

mfjarvis:


coordinator/tests/services/translation/common.spec.ts
```diff
@@ -105,3 +106,37 @@ describe("wrapInOperationOutcome", () => {
     expect(result.issue[0].code).toEqual("invalid")
   })
 })
+
+describe("date time conversion returns UTC timestamp", () => {
+  test("when no timezone present and local time is GMT", () => {
+    const timestamp = convertIsoStringToDateTime("2020-01-21T11:15:30.000")
+    expect(timestamp._attributes.value).toEqual("20200121111530")
```
ri-ch, line 113: What is this date time format? Just a serialized version of the above?

mfjarvis: It's the format specified in the MiM for date times

ri-ch: Ugh

PR #123, opened 2020-08-13
update OperationOutcome to be an interface, define select values for …

dan-jones-iw:
…code and severity as defined in the MiM.


coordinator/src/services/translation/common.ts
```diff
@@ -87,12 +87,14 @@ export function onlyElement<T>(previousValue: T, currentValue: T, currentIndex:
...
 export function wrapInOperationOutcome(message: SpineDirectResponse): fhir.OperationOutcome {
-  const severity = message.statusCode <= 299 ? "information" : "error"
-  const code = message.statusCode <= 299 ? "informational" : "invalid"
-  const operationOutcomeIssue = new fhir.OperationOutcomeIssue(severity, code)
-  operationOutcomeIssue.diagnostics = message.body
-
-  const response = new fhir.OperationOutcome()
-  response.issue = [operationOutcomeIssue]
-  return response
+  const operationOutcomeIssue: fhir.OperationOutcomeIssue = {
```
ri-ch, line 90: We dont need the type here. Given the they type is already defined for the `issue` property on OperationOutcome, if its wrong the compiler will complain about it there

coordinator/src/services/translation/demographics.ts
```diff
@@ -23,7 +23,6 @@ export function convertName(fhirHumanName: fhir.HumanName): core.Name {
 function convertNameUse(fhirNameUse: string) {
   switch (fhirNameUse) {
   case "usual":
-    return core.NameUse.USUAL
```
ri-ch, line 26: This looks like a behaviour change rather than refactoring / code improvements, it would be better if this was in a separate PR

coordinator/src/routes/util.ts
```diff
@@ -41,21 +41,23 @@ function toFhirError(validation: Array<ValidationError>): OperationOutcome {
     * v.operationOutcomeCode: from the [IssueType ValueSet](https://www.hl7.org/fhir/valueset-issue-type.html)
     * v.apiErrorCode: Our own code defined for each particular error. Refer to OAS.
   */
+  const mapValidationErrorToOperationOutcomeIssue = (ve: ValidationError) => {
+    return {
```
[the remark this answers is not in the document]
dan-jones-iw: will change :+1:

PR #128, opened 2020-08-18
Aea 410 strict ts

dan-jones-iw:
3 separate PRs are in one here because i keep pushing changes before they are merged. sorry about the amount of changes, happy to walk through it. if anyone has suggestions on how to break up the code into separate PRs im happy to do that too. 

in summary - changed the fields in our model of MedicationRequest to be consistent with the MiM or the IG in terms of if they should be present or not. as well as that the MedicationRequest model is an interface, and so ive had to change various parts of the codebase to reflect that. happy to go into more detail in a call.


coordinator/tests/services/translation/common/getResourcesOfType.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  test("getPatient", () => {
+    const patient = getPatient(bundle)
+
+    expect(patient.resourceType).toBe("Patient")
+  })
+
+  test("getProvenance", () => {
+    const provenances = getProvenances(bundle)
+
+    provenances.map((provenance) => expect(provenance.resourceType).toBe("Pro"))
```
booshibooshi, line 26: ResourceType should be "Provenance", and the bundle here should be fhirMessageSigned from TestResources

booshibooshi: Before the refactor we were asserting the length of the array returned, maybe we should add this check back in to ensure we're not getting any false positives for tests as in this case

booshibooshi (changes requested):

coordinator/tests/services/translation/common/getResourcesOfType.spec.ts
dan-jones-iw: ahh yeah this is something i left and never came back to, thanks for spotting!

dan-jones-iw: maybe we should have the function throw an error if it doesnt bring back a resource? can chat about it

booshibooshi: discussed and agreed that moving the expectations of number of entries returned for a resource type made sense to be asserted during validation

booshibooshi (approved):

PR #132, opened 2020-08-25
AEA-524 Add human readable display for siganture fragments

booshibooshi:


coordinator/src/services/translation/signing.ts
```diff
@@ -0,0 +1,147 @@
...
+    prescriberOrg: prescriberOrg,
+    prescriberAddress: prescriberAddress,
+    medication: medication
+  }
+
+  return display
+}
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+function joinCoreText(seperator: string, ...values: any[]) : string {
```
dan-jones-iw, line 116: separator

dan-jones-iw: seems like the union type `Array<core.Text | Array<core.Text>>` can be used to remove the eslint disable

dan-jones-iw: not super nice to look at tho, might be a better way of handling it

coordinator/src/services/translation/signing.ts
```diff
@@ -0,0 +1,147 @@
...
+  const medication = fragments.pertinentLineItem
+    .map((medication) => {
+      return new DisplayMedication({
+        name: medication.product.manufacturedProduct.manufacturedRequestedMaterial.code._attributes.displayName,
+        quantity: medication.component.lineItemQuantity.quantity._attributes.value,
+        dose: medication.pertinentInformation2.pertinentDosageInstructions.value
+      })
+    })
+
+  const display: Display = {
```
dan-jones-iw, line 98: return type on function means we can return the object without creating local variable - same on line 11

coordinator/src/services/translation/signing.ts
```diff
@@ -0,0 +1,147 @@
...
+    return text !== undefined ? text._text : ""
+  }
+}
+
+function isCoreTextArray(value: core.Text | core.Text[]) : value is core.Text[] {
+  const valueArray = value as core.Text[]
+  return valueArray !== undefined && valueArray.length > 0
+}
+
+function getGenderFromCode(code: codes.SexCode): string {
```
dan-jones-iw, line 134: already have this function defined in demographics.ts on line 90 - maybe a good idea to collect these into the common.ts file?

coordinator/src/services/translation/signing.ts
```diff
@@ -0,0 +1,147 @@
...
+function joinCoreText(seperator: string, ...values: any[]) : string {
+  return values.map(unpackCoreText).filter(Boolean).join(seperator)
+}
+
+function unpackCoreText(value: core.Text | core.Text[]) : string {
+  if (isCoreTextArray(value)) {
+    return value.map(v => v._text).join(" ")
+  } else {
+    const text = value as core.Text
+    return text !== undefined ? text._text : ""
```
dan-jones-iw, line 125: if text is null or an empty object it will still try and access `text._text` which will throw an error - what you think about just writing `return text ? text._text : ""`?

coordinator/src/model/signing.ts
```diff
@@ -0,0 +1,36 @@
+import * as core from "./hl7-v3-datatypes-core"
+import {GlobalIdentifier} from "./hl7-v3-datatypes-codes"
+import {AgentPerson} from "./hl7-v3-people-places"
+import * as prescriptions from "./hl7-v3-prescriptions"
+
+export class Fragments {
```
ri-ch, line 6: Do these need to classes? Given theres no logic they could be interfaces

ri-ch (commented):
We should probably update the spec to include these changes aswell

coordinator/src/services/translation/signing.ts
booshibooshi: this is actually the reverse of this function back from hl7v3 to fhir. It's very specific to the signing use-case so it made more sense to leave it here and not to pollute fhir to hl7v3 mappings with the reverse function when it's not used as all the other mappings are

dan-jones-iw: ahh of course - my bad

booshibooshi:
> We should probably update the spec to include these changes aswell

Are the changes to the models sufficient to cover these? As these get pulled into the spec. Or are there additional things we'd like to cover?

ri-ch (approved):

PR #137, opened 2020-08-28
Aea 539 cost centre mappings

conner-jowitt:


coordinator/src/model/fhir-resources.ts
```diff
@@ -120,11 +121,19 @@ export class PractitionerRole extends Resource {
...
-  organization?: Reference<Organization>
+  organization?: PractitionerRoleOrganization
   code?: Array<CodeableConcept>
+  healthcareService?: Array<Reference<HealthcareService>>
   telecom: Array<ContactPoint>
 }
 
+export type PractitionerRoleOrganization = PractitionerRoleOrganizationIdentifier | Reference<Organization>
+
+export interface PractitionerRoleOrganizationIdentifier {
```
mfjarvis, line 132: I don't think this is unique to PractitionerRole. Any Reference can refer to its target either using the reference field or the identifier field. I'd rather we made this generic so we can use it elsewhere (e.g. nominated pharmacy).

mfjarvis: Also I'm pretty sure identifier shouldn't be an array.

coordinator/src/services/translation/practitioner.ts
```diff
@@ -46,14 +46,41 @@ export function convertResponsibleParty(
...
+}
+
+function createMiniumumOrganization(display: string, identifier: Array<fhir.Identifier>){
+  return {
+    name: display,
+    identifier,
+    type: [{
+      coding: [{
+        system: "https://fhir.nhs.uk/R4/CodeSystem/organisation-role",
+        code: "RO197"
```
mfjarvis, line 77: I've had a think about this and I think it's a bad idea to set the type. Any Organization can be referred to by its identifier and that doesn't mean it's a trust. It's only in the particular case of PractitionerRole mappings that we want to treat it as such. If we start using this method in other places we're going to run into issues

coordinator/src/model/fhir-resources.ts
conner-jowitt: I thought so too, but then the example on this page: https://nhsd-confluence.digital.nhs.uk/pages/viewpage.action?pageId=132775978 has an array, and then this page: https://simplifier.net/guide/NHSDigitalSpine/Spine-HealthcareService shows a cardinality of 0...*

I'll change the types to be generic

coordinator/src/services/translation/organization.ts
```diff
@@ -1,48 +1,125 @@
...
+
+  return hl7V3Organization
+}
 
-  if (fhirOrganization.type !== undefined) {
-    const organizationTypeCoding = getCodeableConceptCodingForSystem(fhirOrganization.type, "https://fhir.nhs.uk/R4/CodeSystem/organisation-type")
-    hl7V3Organization.code = new codes.OrganizationTypeCode(organizationTypeCoding.code)
+function convertRepresentedOrganization(fhirOrganization: fhir.Organization, fhirHealthcareService: fhir.HealthcareService, fhirBundle: fhir.Bundle) {
+  const organizationTypeCoding = getCodeableConceptCodingForSystemOrNull(fhirOrganization.type, "https://fhir.nhs.uk/R4/CodeSystem/organisation-role")
+  const representedOrganization = (organizationTypeCoding?.code === NHS_TRUST_CODE) ? new CostCentreHealthcareService(fhirHealthcareService, fhirBundle) : new CostCentreOrganization(fhirOrganization)
```
mfjarvis, line 35: Change the condition to be `!organizationTypeCoding || organizationTypeCoding.code === NHS_TRUST_CODE`. See my comment on `resolveOrganizationReference()`.

mfjarvis (approved):

PR #148, opened 2020-09-05
Aea 554 Integrate with common pipelines

booshibooshi:


azure/azure-release-pipeline.yml
```diff
@@ -22,12 +22,37 @@ pool:
...
     short_service_name: ${{ variables.short_service_name }}
     service_base_path: ${{ variables.service_base_path }}
     product_display_name: ${{ variables.product_display_name }}
     product_description: ${{ variables.product_description }}
+    service_base_path_pr: none
+    spec_file: ${{ variables.spec_file }}
+    apigee_deployments:
+      - environment: internal-dev
+        proxy_path: live
+      - environment: internal-qa-sandbox
```
mfjarvis, line 41: Should we deploy to internal-dev-sandbox here as well?

azure/azure-release-template.yml
```diff
@@ -0,0 +1,128 @@
...
+            exit 0
+          fi
+          
+          if [[ $SERVICE_ARTIFACT_NAME == v* ]]; then
+            export BUILD_VERSION=`echo $SERVICE_ARTIFACT_NAME | grep -o "v[0-9]\+\.[0-9]\+\.[0-9]\+-[[:alpha:]]\+" | tail -1`
+            export SERVICE_BASE_PATH="${{ parameters.service_base_path }}"
+          else
+            export BUILD_VERSION="${{ parameters.service_base_path_pr }}"
+            export SERVICE_BASE_PATH="${{ parameters.service_base_path_pr }}"
+          fi
```
mfjarvis, line 51: Is there a way we could end up with no path here?

mfjarvis (approved):

azure/azure-release-template.yml
mfjarvis: I.e. not a PR and artifact name doesn't match the pattern

azure/azure-release-pipeline.yml
booshibooshi: Could look into this now the environments are split, I believe there's still a race condition here in internal-dev, internal-dev-sandbox testing but will raise a ticket for investigation

azure/azure-release-template.yml
booshibooshi: The logic to set the service_base_path_pr is in our repo copied over from utils, so this will never be blank but will default incorrectly to "electronic-prescriptions-pr- " if triggered from a branch, so we're not supporting branch deploys. The logic to use this is based on a non-tag build which is all PRs as tags are created on push to master.

PR #161, opened 2020-09-15
AEA-579 refactor in preparation for sync-wrap integration

booshibooshi:


proxies/live/apiproxy/proxies/default.xml
```diff
@@ -38,6 +38,10 @@
       (proxy.pathsuffix MatchesPath "/_ping") and ((request.verb = "GET") or (request.verb = "HEAD"))
     </Condition>
   </RouteRule>
+  <RouteRule name="sync-wrap">
+    <Condition>request.verb == "GET" and request.header.prefer != "respond-async" and request.header.x-sync-wrapped != "true"</Condition>
```
mfjarvis, line 42: I think we need this to be POST, not GET

mfjarvis: Might also need to filter to just the $process-message endpoint, as that's the one which will respond asynchronously

booshibooshi: Yeah had some questions on this as there's some conflicting documentation for using it. I just wanted to test out communications with sync-wrap initially, but should move it over to process message

booshibooshi: Removed to keep refactoring separate

coordinator/src/routes/prescription/post-send-message.ts
```diff
@@ -1,8 +1,8 @@
-import * as translator from "../../services/translation/translation-service"
-import {Bundle} from "../../model/fhir-resources"
-import {defaultRequestHandler} from "../../services/spine-communication"
+import * as translator from "../../services/translation"
```
dan-jones-iw, line 1: is it possible to rename this file since its no longer `$send` its now `$process-message`.

dan-jones-iw (approved):

PR #167, opened 2020-09-17
Aea 599 translate cancellation messages

conner-jowitt:


coordinator/src/model/fhir-resources.ts
```diff
@@ -43,6 +43,7 @@ export interface MedicationRequest extends Resource {
   dosageInstruction: Array<Dosage>
   dispenseRequest: MedicationRequestDispenseRequest
   extension: Array<IdentifierExtension | ReferenceExtension<PractitionerRole> | CodingExtension | CodeableConceptExtension | RepeatInformationExtension>
+  statusReason?: Array<CodeableConcept>
```
booshibooshi, line 46: Noticed when adding example messages that this fails fhir validation from the tool in our repo as it is expecting an object not an array. The same goes for practitionerRole healthcareService.identifier. I've made some changes in the example messages PR to accomodate, but wanted to flag that this has changed as it does not match the IG

booshibooshi (approved):

PR #173, opened 2020-09-21
Handle some extra cases when translating additional instructions

mfjarvis:


coordinator/src/services/translation/line-item.ts
```diff
@@ -57,27 +59,32 @@ export function convertPrescriptionEndorsements(fhirMedicationRequest: fhir.Medi
...
-  const patientInstructionStr = patientInstruction ? patientInstruction : ""
 
-  const additionalInstructionsValue = `${patientInfoStr}${controlledDrugStr}${patientInstructionStr}`
-  const hl7V3AdditionalInstructions = new prescriptions.AdditionalInstructions(additionalInstructionsValue)
+  const additionalInstructionsValueObj = {} as ElementCompact
+  if (patientInfoText?.length) {
+    additionalInstructionsValueObj.patientInfo = patientInfoText
+  }
+  additionalInstructionsValueObj._text = [controlledDrugWordsWithPrefix, patientInstruction].filter(Boolean).join("\n")
+  const additionalInstructionsValueStr = js2xml(additionalInstructionsValueObj, {compact: true})
```
dan-jones-iw, line 81: does the js2xml maintain the order of elements as we had in the previous version of translation?

dan-jones-iw: and is the ordering important

coordinator/src/model/fhir-resources.ts
```diff
@@ -250,16 +250,19 @@ export interface CommunicationRequest extends Resource {
...
   subject: Reference<Patient>
-  payload: Array<ContentReference | ContentString>
+  payload: Array<ContentString | ContentAttachment | ContentReference>
 }
 
 export interface ContentString {
   contentString: string
 }
 
+export interface ContentAttachment {
```
dan-jones-iw, line 260: delete this and update validations to assure there is a content string

booshibooshi (approved):

PR #177, opened 2020-09-22
AEA-628: Add realistic sandbox responses

mfjarvis:
…alistic.


coordinator/src/services/spine-communication.ts
```diff
@@ -70,18 +70,36 @@ export class SandboxRequestHandler implements RequestHandler {
     }
   }
 
-  async poll(path: string): Promise<SpineResponse<string | OperationOutcome>> {
+  async poll(path: string): Promise<SpineResponse<OperationOutcome>> {
     if (path === this.parentPrescriptionPollingId) {
-      //TODO - add realistic response
+      const parentPrescriptionOperationOutcome = {
```
dan-jones-iw, line 75: instead of `const parent...Outcome = {...} as OperationOutcome`, think its more standard to do `const parent...Outcome: OperationOutcome = {...}`

coordinator/src/services/spine-communication.ts
```diff
@@ -70,18 +70,36 @@ export class SandboxRequestHandler implements RequestHandler {
...
+        ]
+      } as OperationOutcome
       return {
         statusCode: 200,
-        body: "Prescription Message Sent"
+        body: parentPrescriptionOperationOutcome
       }
     } else if (path === this.cancellationPollingId) {
-      //TODO - add realistic response
+      const cancellationRequestOperationOutcome = {
```
dan-jones-iw, line 90: as above

dan-jones-iw (approved):

dan-jones-iw (approved):

PR #191, opened 2020-10-01
Add code scanning workflow

mfjarvis:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+name: "CodeQL"
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+  schedule:
+    - cron: '0 4 * * 6'
```
mfjarvis, line 15: This was the default GitHub gave me - 04:00 on Saturday. Not sure if we'd rather run it daily?

mfjarvis: Having said that, it also runs whenever we merge to master, so weekly should be fine.

booshibooshi (approved):

PR #205, opened 2020-10-05
Aea 664 add secondary care community example

booshibooshi:


README.md
```diff
@@ -95,25 +95,26 @@ npm t
 ```
 
 #### End-to-end tests
-In order for tests under the make target `test-integration-coordinator` to work, you must have built and be running the coordindator locally. In a seperate shell run:
+New examples can be added in the relevant directory under `models/primary-care|secondary-care`
```
conner-jowitt, line 98: Should this be `models/examples/primary-care|secondary-care`?

booshibooshi: updated

README.md
```diff
@@ -95,25 +95,26 @@ npm t
...
-```
-make build
-make run-coordinator
-```
+Following the convention:
 
-Once the coordinator is up and displaying the port number, in another shell run:
+`{number}-{endpoint}-{request|response}-{?operation}-{status}.{ext}`
+
+Number is a way to group requests and responses together in each directory, for example in the below
```
conner-jowitt, line 104: Currently, all the examples have a number of 1, is this correct? For example, in `homecare/acute/no-nominated-pharmacy` there are 6 messages with a number of 1

booshibooshi: yep, it's a way to group examples together, i.e. a convert response for a send request, or prepare and send requests which are the same except for provenance, just we only have one example group for each at the moment

conner-jowitt: ahh ok, think I understand. So the number isn't used in the smoke tests at all? I'll have to wait till I get there and see what it looks like :+1:

scripts/set_env_vars.sh
```diff
@@ -21,8 +21,9 @@ export PACT_BROKER_BASIC_AUTH_USERNAME=
...
 export PACT_CONSUMER=nhsd-apim-eps-test-client
 export PACT_PROVIDER=nhsd-apim-eps-sandbox
-export APIGEE_ENVIRONMENT=internal-dev-sandbox
 export APIGEE_ACCESS_TOKEN=
-export SERVICE_BASE_PATH=electronic-prescriptions-pr-171
-export BUILD_VERSION=AEA-555_refactor-examples+1
-export COMMIT_SHA=8f70bfd1976924c3533bc6f5977d075786e873f8
\ No newline at end of file
+export APIGEE_ENVIRONMENT=internal-dev-sandbox
+export SERVICE_BASE_PATH=electronic-prescriptions-pr-189
```
conner-jowitt, line 26: If this is a placeholder to show how to run pacts, is it worth adding a section in the `running tests` section of the readme explaining that this needs setting?

booshibooshi: that would probably be helpful, i'll add this comment to the ACs for the ticket i'm working on now

tests/e2e/pact/models/cases/process-case.ts
```diff
@@ -0,0 +1,17 @@
+import {Bundle} from "../fhir/fhir-resources"
```
conner-jowitt, line 1: These classes share some common fields and the constructors are all quite similar, could they be cleaned up through inheritance?. If this ticket is needed urgently might be worth raising another ticket to do it?

booshibooshi: Yep a refactor here would be ideal in another ticket

tests/e2e/pact/services/convert-example-fetcher.ts
```diff
@@ -0,0 +1,47 @@
+import path from "path"
+import { ConvertCase } from "../models/cases/convert-case"
+import { exampleFiles } from "./example-files-fetcher"
+
+const examplesRootPath = "../resources/parent-prescription"
+
+const convertResponseFiles = exampleFiles.filter(exampleFile => exampleFile.isResponse && exampleFile.endpoint === "convert")
+
+const convertRequestFiles = exampleFiles.filter(exampleFile =>
```
conner-jowitt, line 9: This filter is confusing to read at first. Not urgent but could be cleaned up in a separate ticket to make it more readable?

tests/e2e/pact/services/convert-example-fetcher.ts
```diff
@@ -0,0 +1,47 @@
...
+	exampleFile.isRequest
+	&& convertResponseFiles.some(convertResponseFile => 
+			convertResponseFile.dir === exampleFile.dir
+			&& convertResponseFile.operation === exampleFile.operation
+			&& convertResponseFile.number === exampleFile.number))
+
+const conventionBasedConvertExamples: ConvertCase[] = convertResponseFiles.map(convertResponseFile => new ConvertCase(
```
conner-jowitt, line 16: Same here, could do with cleaning up these arguments with some helper functions to make it more readable?

tests/e2e/pact/services/convert-example-fetcher.ts
booshibooshi: Sounds good

tests/e2e/pact/services/prepare-example-fetcher.ts
```diff
@@ -0,0 +1,35 @@
+import path from "path"
+import { PrepareCase } from "../models/cases/prepare-case"
+import { exampleFiles } from "./example-files-fetcher"
+
+const examplesRootPath = "../resources/parent-prescription"
+
+const prepareResponseFiles = exampleFiles.filter(exampleFile => exampleFile.isResponse && exampleFile.endpoint === "prepare")
+
+const prepareRequestFiles = exampleFiles.filter(exampleFile =>
```
conner-jowitt, line 9: Won't comment twice, but this file needs the same changes as `convert-example-fetcher`

tests/e2e/pact/services/process-example-fetcher.ts
```diff
@@ -0,0 +1,22 @@
+import path from "path"
+import { ProcessCase } from "../models/cases/process-case"
+import { exampleFiles } from "./example-files-fetcher"
+
+const examplesRootPath = "../resources/parent-prescription"
+
+const processRequestFiles = exampleFiles.filter(exampleFile => exampleFile.isRequest && exampleFile.endpoint === "process")
+
+const conventionBasedProcessExamples: ProcessCase[] = processRequestFiles.map(processRequestFile => new ProcessCase(
```
conner-jowitt, line 9: Same as before, helper function for arguments

conner-jowitt (approved):

PR #212, opened 2020-10-08
Readme updates

dan-jones-iw:


coordinator/README.md
```diff
@@ -11,11 +11,27 @@ npm install
...
+
 ## Deployment
 
 Redeploy the API Proxy. See the main [README.md](../README.md).
 
 ## Endpoints
 
-- [ ] POST `/ConvertFullMessage`
-- [ ] POST `/ConvertSignatureFragments`
+Endpoints are found in `/src/routes`.
```
harryhmc, line 31: You've not mentioned the `/src` directory in the file paths in the Directories section but you have here. Maybe just write `/routes` for consistency with above?

coordinator/README.md
```diff
@@ -11,11 +11,27 @@ npm install
 npm run serve
 ```
 
+### Directories
+- `/models` Typescript interface/class definitions
+- `/resources` mustache template files for xml responses
+- `/routes` API endpoint definitions
```
harryhmc, line 17: Do `/routes` subdirectories need to be added here, or is that a level of detail we can comfortably leave out?

harryhmc (commented):
You'll need to update the `make release` command for this repo -- it's still referencing the terraform directory which is causing the build pipeline to fail

coordinator/README.md
dan-jones-iw: just did it for consistency, gone into more detail below, can delete if u think its pointless

coordinator/README.md
dan-jones-iw: was thinking about putting `/src` in the directories section tbh, maybe that would make this better?

harryhmc: yh that looks good

harryhmc (approved):

harryhmc (approved):

harryhmc (approved):

PR #219, opened 2020-10-15
Aea 729 Map professional and prescribing codes

booshibooshi:


coordinator/src/models/hl7-v3/hl7-v3-datatypes-codes.ts
```diff
@@ -160,21 +160,33 @@ export class NhsNumber extends Identifier<"2.16.840.1.113883.2.1.4.1"> {
...
-export class SdsUniqueIdentifier extends Identifier<"1.2.826.0.1285.0.2.0.65"> {
+export class AgentPersonIdCode extends Identifier<"1.2.826.0.1285.0.2.1.54"> {
   constructor(extension: string) {
-    super("1.2.826.0.1285.0.2.0.65", extension)
+    super("1.2.826.0.1285.0.2.1.54", extension)
   }
 }
 
-export class SdsRoleProfileIdentifier extends Identifier<"1.2.826.0.1285.0.2.0.67"> {
+export class PrescribingCode extends AgentPersonIdCode {
```
mfjarvis, line 169: Is there any reason to have these two subclasses given they're exactly the same? I'd be tempted to replace all uses with the superclass.

coordinator/src/models/hl7-v3/hl7-v3-people-places.ts
```diff
@@ -140,7 +140,7 @@ export class PatientPerson implements ElementCompact {
     playedProviderPatient: ProviderPatient
 }
 
-export type PrescriptionAuthorId = codes.BsaPrescribingIdentifier | codes.SdsUniqueIdentifier
+export type PrescriptionAuthorId = codes.ProfessionalCode | codes.PrescribingCode
```
mfjarvis, line 143: This type is basically AgentPersonIdCode. Could replace it with that?

coordinator/src/services/translation/prescription/practitioner.ts
```diff
@@ -40,7 +41,9 @@ export function convertResponsibleParty(
   fhirBundle: fhir.Bundle,
   fhirMedicationRequest: fhir.MedicationRequest,
   isCancellation: boolean,
-  convertPractitionerRoleFn = convertPractitionerRole
+  convertPractitionerRoleFn = convertPractitionerRole,
```
mfjarvis, line 44: Besides the id function, none of these defaults ever seem to be overridden, even in tests. Might be nice to get rid of these parameters

mfjarvis (approved):

coordinator/src/models/hl7-v3/hl7-v3-datatypes-codes.ts
booshibooshi: They have meaning in the domain, intention was to expose these conceptually as there's logic to fall back from prescribing to professional code, yes they have the same oid, but the logic to select a prescribing code vs professional is clearer when they are different classes

coordinator/src/services/translation/prescription/practitioner.ts
booshibooshi: Agreed

PR #220, opened 2020-10-16
AEA-744: Gated releases to external environments

mfjarvis:


azure/azure-external-release-pipeline.yml
```diff
@@ -0,0 +1,48 @@
...
+    short_service_name: ${{ variables.short_service_name }}
+    service_base_path: ${{ variables.service_base_path }}
+    product_display_name: ${{ variables.product_display_name }}
+    product_description: ${{ variables.product_description }}
+    service_base_path_pr: none
+    spec_file: ${{ variables.spec_file }}
+    apigee_deployments:
+      - environment: sandbox
+        proxy_path: sandbox
+        depends_on:
```
booshibooshi, line 35: Do these need updating to depends_on: [] now that internal_* environments are in the internal pipeline? If not specified then these get depends_on defaults from common pipeline, [] overrides

mfjarvis: Yeah possibly. Trying the new pipelines out in Azure DevOps now

booshibooshi (approved):

PR #221, opened 2020-10-16
AEA-748 Update electronic-prescription-service-api.yaml

tjhmoore:


SachinPath (approved):

SachinPath:
The change is in line with the PDS FHIR API specs.

PR #223, opened 2020-10-19
AEA-744: Gated releases attempt two

mfjarvis:


azure/azure-external-release-pipeline.yml
```diff
@@ -9,6 +9,9 @@ resources:
       type: github
       name: NHSDigital/api-management-utils
       endpoint: NHSDigital
+  pipelines:
+    - pipeline: build_pipeline
+      source: "Electronic-Prescription-Service-Build"
```
booshibooshi, line 15: Does this need the trigger filter to be consistent with internal release?

```
trigger:
      branches:
        include:
          - refs/tags/v*
```

So only tagged releases from master trigger?

booshibooshi (approved):

PR #226, opened 2020-10-19
AEA-740: Update the rules for repeat dispensing prescriptions

mfjarvis:


coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -165,7 +176,7 @@ function verifyMessageIsResource(message: unknown): message is fhir.Resource {
 
 function verifyResourceTypeIsBundle(resource: unknown): resource is fhir.Bundle {
   return verifyMessageIsResource(resource)
-        && resource.resourceType === "Bundle"
+    && resource.resourceType === "Bundle"
```
harryhmc, line 179: verifying the resourceType is Bundle is presumably the easier check -- could you flip the conditions around the `&&` to make use of short-circuiting?

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -126,6 +127,15 @@ export function verifyPrescriptionBundle(bundle: fhir.Bundle): Array<errors.Vali
         "https://fhir.nhs.uk/R4/StructureDefinition/Extension-DM-ResponsiblePractitioner",
         "MedicationRequest.extension"
       )
+    ),
+    (medicationRequests: Array<fhir.MedicationRequest>) => verifyValueIdenticalForAllMedicationRequests(
```
harryhmc, line 131: is it worth adding a test to `request-validator.spec.ts` around this being enforced?

harryhmc: also i know you've only added one of the validators here but there has to be a more concise way of creating this array

harryhmc (approved):

coordinator/src/services/validation/bundle-validator.ts
mfjarvis: Didn't bother adding a test as we already have tests for the consistency check, and the only difference here is which field is being considered. As for creating the array, Conner's neatened it up on branch AEA-628-improve-request-validator.

PR #232, opened 2020-10-23
Aea 755 tel mappings format

conner-jowitt:


coordinator/src/models/hl7-v3/hl7-v3-datatypes-core.ts
```diff
@@ -173,6 +173,12 @@ export class Telecom implements ElementCompact {
   }
 
   constructor(use: TelecomUse, value: string) {
+    if (value) {
+      value = value.replace(/\s/g, "")
+      if (!value.startsWith("tel:")) {
+        value = `tel:${value}`
```
mfjarvis, line 179: Do we need to handle email, fax and textphone? PDS does, but maybe we can say these are unsupported for EPS?

conner-jowitt: Not sure, I'll double check with Chris to see if it's required

mfjarvis (dismissed):

mfjarvis (approved):

PR #233, opened 2020-10-27
AEA-582: Return signature fragments from the $prepare endpoint

mfjarvis:


coordinator/src/services/translation/index.ts
```diff
@@ -104,9 +103,10 @@ function createParametersDisplay(fragmentsToDisplay: Display): string {
   return Buffer.from(Mustache.render(displayTemplate, fragmentsToDisplay)).toString("base64")
 }
 
-function createParameters(base64Payload: string, base64Display: string): fhir.Parameters {
+function createParameters(base64Fragments: string, base64Digest: string, base64Display: string): fhir.Parameters {
   const parameters: Array<fhir.Parameter> = []
-  parameters.push({name: "payload", valueString: base64Payload})
+  parameters.push({name: "fragments", valueString: base64Fragments})
```
mfjarvis, line 108: The clearest name for this parameter is `fragments`, but calling it `payload` would mean the Signing Service would accept it unchanged

coordinator/src/services/translation/index.ts
```diff
@@ -104,9 +103,10 @@ function createParametersDisplay(fragmentsToDisplay: Display): string {
...
+  parameters.push({name: "digest", valueString: base64Digest})
```
mfjarvis, line 109: The digest is no longer required by the Signing Service, but I've left it in case it's needed by the supplier (in order to build the XML-DSig) or by the Remote Signing Service in future

conner-jowitt (approved):

PR #238, opened 2020-10-29
AEA-628: Add alternative way of verifying consistency across MedicationRequests

mfjarvis:


coordinator/src/services/validation/fhir-path.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export function applyFhirPath(bundle: Bundle, input: Array<unknown>, path: string): Array<unknown> {
+  const pathElements = splitFhirPath(path)
+  return pathElements.reduce(
+    (output, pathElement) => applyFhirPathElement(bundle, output, pathElement),
+    input
+  )
+}
+
+function splitFhirPath(path: string) {
```
mfjarvis, line 17: This function takes a FHIR path, e.g.:
`entry.resource.ofType(MedicationRequest).dispenseRequest.extension("https://fhir.nhs.uk/R4/StructureDefinition/Extension-performerSiteType")`
and returns the individual elements, e.g.:
* `entry`
* `resource`
* `ofType(MedicationRequest)`
* `dispenseRequest`
* `extension("https://fhir.nhs.uk/R4/StructureDefinition/Extension-performerSiteType")`

Note we can't just split on `.` because the `extension()` elements contain URLs.

This is a fairly naive implementation - it doesn't for example deal with escaped quotes. However we only ever parse FHIR paths which we have written, never ones from a user, so this doesn't have to be bulletproof.

coordinator/tests/services/validation/util.spec.ts
```diff
@@ -0,0 +1,46 @@
+import {getUniqueValues} from "../../../src/services/validation/util"
+
+describe("getUniqueValues returns correct values", () => {
+  test("when input is empty", () => {
```
mfjarvis, line 4: There's probably an argument to be had about whether an empty array contains no unique values, or the single unique value `undefined`. I don't think it matters for the purposes of our FHIR validation, as long as it treats `undefined` as being different to an actual value.

coordinator/src/services/validation/util.ts
```diff
@@ -0,0 +1,14 @@
...
+export function getUniqueValues<T>(allValues: Array<T>): Array<T> {
+  return allValues.reduce(
+    (existingUniqueValues, valueToTest) => {
+      if (existingUniqueValues.every((uniqueValue) => !isDeepStrictEqual(uniqueValue, valueToTest))) {
+        return existingUniqueValues.concat(valueToTest)
+      } else {
+        return existingUniqueValues
+      }
+    },
+    [allValues.shift()]
```
ri-ch, line 12: Presumably, the reason we're popping the first value off the array here is because we need something to seed the reduce function with?

coordinator/src/models/errors/validation-errors.ts
```diff
@@ -13,12 +14,12 @@ export class MedicationRequestValueError implements ValidationError {
   severity = "error" as const
   expression: Array<string>
 
-  constructor(fieldName: string, uniqueFieldValues: Array<string>) {
+  constructor(fieldName: string, uniqueFieldValues: Array<unknown>) {
```
ri-ch, line 17: Can we make this generic instead of Unknown?

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -21,121 +22,46 @@ export function verifyBundle(bundle: fhir.Bundle): Array<errors.ValidationError>
...
-  fieldName: string,
-  fieldAccessor: (resource: fhir.MedicationRequest) => U
-): errors.ValidationError | null {
-  const fieldValues = medicationRequests.map(fieldAccessor)
-  const serializedFieldValues = fieldValues.map(value => LosslessJson.stringify(value))
-  const uniqueFieldValues = new Set(serializedFieldValues)
-  return uniqueFieldValues.size === 1 ? null : new errors.MedicationRequestValueError(fieldName, [...uniqueFieldValues])
+  return fhirPaths
+    .map((fhirPath) => verifyIdenticalForAllMedicationRequests(bundle, medicationRequests, fhirPath))
+    .filter(Boolean)
```
ri-ch, line 47: Is this actually filtering out all non-null values?

coordinator/src/services/validation/fhir-path.ts
mfjarvis: But then we'd have two problems

ri-ch: Very clever 😉 Happy to go either way, but all things being equal I prefer having less code rather than more

coordinator/src/services/validation/bundle-validator.ts
mfjarvis: The tests pass so I'm going to say yes

mfjarvis: I think we use this in other places too

coordinator/src/models/errors/validation-errors.ts
mfjarvis: Yeah makes sense. This is one of the places that `any` would actually be appropriate

coordinator/src/services/validation/fhir-path.ts
```diff
@@ -0,0 +1,61 @@
+import {Bundle, Extension, Reference, Resource} from "../../models/fhir/fhir-resources"
+import {resolveReference} from "../translation/common"
+
+/**
+ * Implementation of simple FHIR paths
+ * See https://www.hl7.org/fhir/fhirpath.html#simple
+ */
+
+export function applyFhirPath(bundle: Bundle, input: Array<unknown>, path: string): Array<unknown> {
```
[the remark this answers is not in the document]
mfjarvis: We could use this with any FHIR object, not just MedicationRequests. For example one of the tests calls it with Bundle.entry

coordinator/src/services/validation/fhir-path.ts
```diff
@@ -0,0 +1,61 @@
...
+      currentPathElement += char
+    }
+  }
+  if (currentPathElement) {
+    pathElements.push(currentPathElement)
+  }
+  return pathElements
+}
+
+function applyFhirPathElement(bundle: Bundle, input: Array<unknown>, pathElement: string): Array<unknown> {
```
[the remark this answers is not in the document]
mfjarvis: Not quite. I was going for "apply one segment of a FHIR path". Is there a clearer way to say that?

coordinator/src/services/validation/fhir-path.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (pathElement === "resolve()") {
+    const references = input as Array<Reference<Resource>>
+    return references.map(i => resolveReference(bundle, i))
+  }
+
+  const records = input as Array<Record<string, unknown>>
+  const extensionMatch = pathElement.match(/extension\("(.*)"\)/)
```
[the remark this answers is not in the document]
mfjarvis: Can do

coordinator/src/services/validation/util.ts
mfjarvis: Yeah that's it

coordinator/src/services/validation/fhir-path.ts
mfjarvis: Nah I was kidding. Didn't occur to me to use regex

mfjarvis: Although I'm going to need to convince myself that that regex does what I want it do. Which is the joy of using regex instead of code :unamused:

ri-ch (approved):

PR #239, opened 2020-10-29
Change validator branch to master

mfjarvis:


mfjarvis:
Looks like this change made it into master via another PR

PR #242, opened 2020-11-03
Update ref ig3 url logic to match existing common parameter

booshibooshi:


mfjarvis (approved):

coordinator/src/services/handlers/spine-handler.ts
```diff
@@ -124,15 +124,15 @@ export class LiveRequestHandler implements RequestHandler {
 
   private getSpineUrlForPrescription() {
     if (this.spineEndpoint.includes("ref")) {
-      return `${SPINE_URL_SCHEME}://${this.spineEndpoint.replace(/demographics/g, "prescriptions")}/${this.spinePath}`
+      return `${SPINE_URL_SCHEME}://${this.spineEndpoint.replace(/msg/g, "prescriptions")}/${this.spinePath}`
```
ri-ch, line 127: This is a little bit hooky still. Is there no way we can just inject the correct URL to start with?

booshibooshi: Yep, we should have our own secret in EPS namespace, just patching to unblock ref for now

PR #247, opened 2020-11-12
Handle spine error responses (ish)

ri-ch:


conner-jowitt (approved):

coordinator/tests/services/spine-communication.spec.ts
```diff
@@ -69,12 +71,64 @@ describe("Spine communication", () => {
...
+    })
+
+    const spineResponse = await requestHandler.send({message: "test", interactionId: "test2"})
+
+    expect(spineResponse.statusCode).toBe(400)
+    expect(isDirect(spineResponse)).toBe(true)
+    expect((spineResponse as SpineDirectResponse<string>).body).toContain("<acknowledgement typeCode=\"AR\">")
+  })
+
+  test("Async success messages returned from spine return a 200 response", async () => {
```
dan-jones-iw, line 108: we dont test sync success?

coordinator/src/services/handlers/spine-handler.ts
```diff
@@ -81,14 +81,20 @@ export class LiveRequestHandler implements RequestHandler {
     }
   }
 
+  private static handlePotentialFailureCaseForDirectResponse(data: string | undefined): SpineDirectResponse<string> {
+    // If the body contains the magic tag, this represents an error and we should return a 400
+    const isSuccess = data && data.includes("acknowledgement typeCode=\"AA\"")
```
dan-jones-iw, line 86: think its fairly normal to use single quotes to avoid having to escape double quotes in strings (and pretty sure we do this in other places)

coordinator/tests/services/spine-communication.spec.ts
```diff
@@ -69,12 +71,64 @@ describe("Spine communication", () => {
...
+        statusText: "OK",
+        responseText: readFileAsString("async_error.xml")
+      })
+    })
+
+    const spineResponse = await requestHandler.send({message: "test", interactionId: "test2"})
+
+    expect(spineResponse.statusCode).toBe(400)
+    expect(isDirect(spineResponse)).toBe(true)
+    expect((spineResponse as SpineDirectResponse<string>).body).toContain("<hl7:acknowledgement typeCode=\"AE\">")
```
dan-jones-iw, line 88: see above

coordinator/tests/services/spine-communication.spec.ts
mfjarvis: For EPS, our success responses will only ever be async. Could add a test anyway but we'd have to make up the message

dan-jones-iw (dismissed):

mfjarvis (approved):

PR #250, opened 2020-11-12
AEA-798: Extra validation

mfjarvis:


coordinator/src/models/errors/validation-errors.ts
```diff
@@ -24,6 +24,25 @@ export class MedicationRequestValueError<T> implements ValidationError {
   }
 }
 
+export class MedicationRequestNumberError implements ValidationError {
+  operationOutcomeCode = "value" as const
+  severity = "error" as const
+  message = `Expected exactly one MedicationRequest in a prescriptionOrderUpdate message`
+  expression = [`Bundle.entry.resource.ofType(MedicationRequest)`]
```
dan-jones-iw, line 31: dont need backtick strings

coordinator/src/models/errors/validation-errors.ts
```diff
@@ -24,6 +24,25 @@ export class MedicationRequestValueError<T> implements ValidationError {
...
+}
+
+export class MedicationRequestMissingValueError implements ValidationError {
+  message: string
```
dan-jones-iw, line 35: kinda dumb that you have to 'define' these again even tho they're already defined in `ValidationError` but guess thats just TS

coordinator/src/services/translation/prescription/demographics.ts
```diff
@@ -72,21 +72,15 @@ export function convertAddress(fhirAddress: fhir.Address, fhirPath: string): cor
...
     hl7V3Address.postalCode = new core.Text(fhirAddress.postalCode)
   }
   return hl7V3Address
 }
 
-function convertAddressUse(fhirAddressUse: string, fhirAddressType: string, fhirPath: string) {
-  if (fhirAddressUse === undefined && fhirAddressType === undefined){
-    return undefined
-  }
-  if (fhirAddressType === "postal") {
```
dan-jones-iw, line 87: why isthis being removed?

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -42,20 +43,53 @@ export function verifyPrescriptionBundle(bundle: fhir.Bundle): Array<errors.Vali
...
 }
 
-function verifyCancellationBundle(bundle: fhir.Bundle): Array<errors.ValidationError> {
-  //TODO - implement cancellation-specific validation
-  bundle
+export function verifyRepeatDispensingPrescription(
+  medicationRequests: Array<fhir.MedicationRequest>
+): Array<errors.ValidationError> {
+  if (getCourseOfTherapyTypeCode(medicationRequests) !== CourseOfTherapyTypeCode.CONTINUOUS_REPEAT_DISPENSING) {
+    return []
```
dan-jones-iw, line 57: should this return an error?

coordinator/src/models/errors/validation-errors.ts
mfjarvis: \`maybe I like backtick strings\`

coordinator/src/services/translation/prescription/demographics.ts
mfjarvis: Because it doesn't match the way PDS has mapped address types. Their API only looks at the address use, not the address type. We're making our API consistent with that

coordinator/src/services/validation/bundle-validator.ts
mfjarvis: No, this just means the prescription is not a repeat dispensing one, so none of the validation applies

coordinator/src/services/translation/prescription/demographics.ts
dan-jones-iw: okay just making sure its intended :+1:

coordinator/src/services/validation/bundle-validator.ts
booshibooshi: I think based on this question being asked it's probably clearer if the check for the course of therapy type is outside of this method and this method is only called for repeat dispensing medication requests to avoid confusion. It's not obvious from the method name that this can be run against acute prescriptions without raising errors

coordinator/src/models/errors/validation-errors.ts
mfjarvis: Sorted

coordinator/src/services/validation/bundle-validator.ts
mfjarvis: Sorted

booshibooshi (approved):

PR #260, opened 2020-11-20
convert spine errors into display in operation outcome

dan-jones-iw:


Makefile
```diff
@@ -12,7 +12,7 @@ all:
 
 install: install-node install-python install-hooks
 
-build: build-specification build-coordinator build-validator build-proxies
+build: build-specification build-coordinator build-proxies
```
dan-jones-iw, line 15: this change was to allow me to commit changes - dont have `mvn` installed. can we make it so we dont need maven installed to commit things? i didnt make changes to the validator so doesnt seem to make much sense to rebuild it.

coordinator/src/services/translation/common/index.ts
```diff
@@ -173,14 +177,70 @@ export function convertMomentToHl7V3DateTime(dateTime: moment.Moment): core.Time
...
+  const bodyString = message.body.toString()
+  const syncMCCI = SYNC_SPINE_RESPONSE_MCCI_REGEX.exec(bodyString)
+  const asyncMCCI = ASYNC_SPINE_RESPONSE_MCCI_REGEX.exec(bodyString)
+  let codeableConceptArray: Array<fhir.CodeableConcept> = []
+  if (syncMCCI) {
+    codeableConceptArray = translateSyncSpineResponse(syncMCCI[0])
+  } else if (asyncMCCI) {
+    codeableConceptArray = translateAsyncSpineResponse(asyncMCCI[0])
+  } else {
+    throw new Error("Spine response doesnt match")
```
dan-jones-iw, line 233: is this the best way to handle an unexpected spine error response?

coordinator/src/services/translation/common/index.ts
```diff
@@ -173,14 +177,70 @@ export function convertMomentToHl7V3DateTime(dateTime: moment.Moment): core.Time
...
     resourceType: "OperationOutcome",
-    issue: [{
-      code: message.statusCode <= 299 ? "informational" : "invalid",
-      severity: message.statusCode <= 299 ? "information" : "error",
-      diagnostics: message.body?.toString()
-    }]
+    issue: codeableConceptArray.map(codeableConcept => ({
+      code:"invalid",
+      severity: "error",
+      diagnostics: message.body?.toString(),
```
dan-jones-iw, line 241: we can take this out now - do we need diagnostics? the FHIR spec says this should be populated with 'additional diagnostic information'.

Makefile
mfjarvis: Think this has broken the build - it's called by the pipeline as well

mfjarvis: Why not install maven? It takes like 5 seconds

coordinator/src/services/translation/common/index.ts
mfjarvis: I agree we want to get rid of it eventually. I'd be inclined to leave it in for now as will help us look into issues

coordinator/src/services/translation/common/index.ts
mfjarvis: It's _a_ way :shrug:

mfjarvis: (I think it's fine as long as we log out the content of the response somewhere, which I think Axios already does for us)

mfjarvis (approved):

PR #263, opened 2020-11-23
AEA-798: Missing field validation

mfjarvis:


coordinator/src/services/translation/prescription/index.ts
```diff
@@ -184,9 +180,7 @@ function convertPrescriptionPertinentInformation2(
   const pertinentInformation2 = []
 
   for (let i = 0; i < fhirMedicationRequests.length; i++) {
```
ri-ch, line 182: This function is effectively a map over the incoming FHIR medication requests, so we can probably refactor things slightly to avoid the clunky `for` loop and the temporary array:
```typescript
return fhirMedicationRequests.map((medicationRequest, index) => {
    const patientInfoText = index === 0 ? extractText(fhirCommunicationRequests) : []
    const pertinentLineItem = convertMedicationRequestToLineItem(
      medicationRequest,
      repeatNumber,
      patientInfoText
    )
    return new prescriptions.PrescriptionPertinentInformation2(pertinentLineItem)
  })
```

coordinator/src/services/translation/prescription/index.ts
```diff
@@ -166,14 +166,10 @@ function convertDispensingSitePreference(
 function extractText(fhirCommunicationRequests: Array<fhir.CommunicationRequest>): Array<core.Text> {
   return fhirCommunicationRequests
     .flatMap(communicationRequest => communicationRequest.payload)
```
ri-ch, line 168: As this is a flat map, it will already have extracted the content strings out of the underlying payload object and turned them into `Array<ContentString>`, so line 170 is a little confusing.

coordinator/src/services/translation/prescription/index.ts
```diff
@@ -166,14 +166,10 @@ function convertDispensingSitePreference(
...
-    .map(contentString => new core.Text(contentString.contentString))
-}
-
-function isFirstRequestAndCommunicationRequestPresent(
-  request: number,
-  fhirCommunicationRequest: Array<fhir.CommunicationRequest>
-) {
-  return (request == 0 && fhirCommunicationRequest.length > 0)
+    .filter(Boolean)
+    .map(payload => payload.contentString)
```
ri-ch, line 170: As mentioned above, should this be `.map(contentString => contentString.contentString)`? The double level naming is actually a little bit confusing, now that i look at it.

coordinator/src/services/translation/prescription/index.ts
```diff
@@ -184,9 +180,7 @@ function convertPrescriptionPertinentInformation2(
   const pertinentInformation2 = []
 
   for (let i = 0; i < fhirMedicationRequests.length; i++) {
-    const patientInfoText = isFirstRequestAndCommunicationRequestPresent(i, fhirCommunicationRequests)
-      ? extractText(fhirCommunicationRequests)
-      : []
+    const patientInfoText = i === 0 ? extractText(fhirCommunicationRequests) : []
```
ri-ch, line 183: A comment to explain that we only want to set the patient info text for the first item might be useful, or alternatively (to 'self document' the code, create a small 'isFirstMedicationRequest` function.

coordinator/src/services/translation/prescription/organization.ts
```diff
@@ -35,24 +31,30 @@ export function convertOrganizationAndProviderLicense(
...
+  const representedOrganization = isNhsTrust(fhirOrganization)
+    ? new CostCentreHealthcareService(fhirHealthcareService)
+    : new CostCentreOrganization(fhirOrganization)
+  return convertRepresentedOrganizationDetails(representedOrganization, fhirBundle)
+}
+
+function isNhsTrust(fhirOrganization: fhir.Organization) {
   const organizationTypeCoding = getCodeableConceptCodingForSystemOrNull(
-    fhirOrganization.type,
+    fhirOrganization.type || [],
```
ri-ch, line 44: It would be nicer if the handling for an invalid organizationType could be handled inside the `getCodeableConceptCodingForSystemOrNull` function. Then we don't need to leak out the logic into the calling function.

coordinator/src/services/translation/prescription/organization.ts
```diff
@@ -18,14 +19,9 @@ export function convertOrganizationAndProviderLicense(
...
-  const hl7V3Organization = convertRepresentedOrganization(
-    fhirOrganization,
-    fhirHealthcareService,
-    fhirBundle,
-    isCancellation
-  )
+  const hl7V3Organization = convertRepresentedOrganization(fhirOrganization, fhirHealthcareService, fhirBundle)
 
-  if (!isCancellation){
+  if (!isCancellation) {
```
ri-ch, line 24: This is probably a bigger change than we want to do here, but... we don't actually check the value of `isCancellation` until this point, yet it has been passed down through all the levels of code to get here. Given we also pass the bundle which contains the same information (whether or not the message represents a cancellation) it would be nice to remove the redundant flag, and only obtain the value at this point.

ri-ch (commented):
A few comments

coordinator/src/services/translation/prescription/index.ts
mfjarvis: Ah, didn't know about the index parameter. Not my code, but I'll look at changing it

coordinator/src/services/translation/prescription/index.ts
mfjarvis: Yeah maybe the class should be ContentStringPayload or something

coordinator/src/services/translation/prescription/organization.ts
mfjarvis: Good idea. Don't think it's that bit a change, will look at doing it

coordinator/src/services/translation/prescription/organization.ts
mfjarvis: Yeah that's fair I was being lazy

ri-ch (approved):

PR #267, opened 2020-11-25
AEA-789 added smoke tests for unauthorised requests

harryhmc:


tests/e2e/pact/specs/eps-convert.spec.ts
```diff
@@ -35,7 +84,8 @@ jestpact.pactWith(
           uponReceiving: `a request to convert ${desc} message`,
           withRequest: {
             headers: {
-              "Content-Type": "application/fhir+json; fhirVersion=4.0"
+              "Content-Type": "application/fhir+json; fhirVersion=4.0",
+              "Authorization": `Bearer ${process.env.APIGEE_ACCESS_TOKEN}`
```
mfjarvis, line 88: Does this mean an access token will appear in the published pact?

mfjarvis: Not sure if it's actually a problem, but might raise eyebrows if we start sharing pacts with suppliers

harryhmc: hm, good point :thinking: i could revert the changes to the verifier and make a separate verifier for testing auth? would be a bit more work but would avoid this issue

ri-ch (approved):

tests/e2e/pact/specs/eps-auth.spec.ts
```diff
@@ -0,0 +1,78 @@
+import { InteractionObject } from "@pact-foundation/pact"
+import * as jestpact from "jest-pact"
+import supertest from "supertest"
+
+jestpact.pactWith(
+  {
+    spec: 3,
+    consumer: `nhsd-apim-eps-test-client+${process.env.PACT_VERSION}`,
+    provider: `${process.env.PACT_PROVIDER}+${process.env.PACT_VERSION}`,
```
booshibooshi, line 9: This will overwrite the previous pact, effectively replacing all the model example tests with just the 3 invalid auth ones

booshibooshi (changes requested):

tests/e2e/pact/package.json
```diff
@@ -23,6 +23,7 @@
     "@types/jest": "^26.0.0",
     "@types/lossless-json": "^1.0.0",
     "@types/node": "^14.0.13",
+    "@types/puppeteer": "^5.4.0",
```
booshibooshi, line 26: is this dep used?

booshibooshi (approved):

tests/e2e/pact/package.json
ri-ch: It is not. Good spot

booshibooshi (approved):

PR #269, opened 2020-12-02
undoing certificate change

conner-jowitt:


mfjarvis:
Spine communication is now working with the new certs so no need to revert.

PR #274, opened 2020-12-08
Set auth bearer token in eps spec through request-filter to enable invalid auth tests

booshibooshi:


harryhmc (approved):
LGTM, will merge :+1:

PR #276, opened 2020-12-08
Translate prescription 1 - Add moxifloxican example

booshibooshi:


ri-ch (dismissed):

ri-ch (dismissed):

ChrisTatts2 (commented):
Have sent the email to Craig as the Effective date doesnt look correct to me when we convert it into XML Also within the message we have the Role as R8000 when it should be R8003 though this isnt a showstopper.

ri-ch (approved):

PR #279, opened 2020-12-11
Improve smoke tests

booshibooshi:
.


azure/azure-release-template.yml
```diff
@@ -70,29 +71,27 @@ extends:
             if [ "$APIGEE_ENVIRONMENT" == "int" ]
             then
               export IDP_URL="https://nhsd-apim-testing-$APIGEE_ENVIRONMENT-ns.herokuapp.com"
+              docker pull artronics/nhsd-login-docker:latest > /dev/null 2>&1
```
mfjarvis, line 74: Would it be better if we didn't redirect stderr to stdout? I think that way we'd still be able to see if something goes wrong

mfjarvis: Also could move this to line 70 to avoid duplicating it

azure/azure-release-template.yml
```diff
@@ -56,12 +56,13 @@ extends:
           fi
 
           export APIGEE_ENVIRONMENT=$(APIGEE_ENVIRONMENT)
-
           export PACT_BROKER_URL=$(PACT_BROKER_URL)
           export PACT_BROKER_BASIC_AUTH_USERNAME=$(PACT_BROKER_BASIC_AUTH_USERNAME)
           export PACT_BROKER_BASIC_AUTH_PASSWORD=$(PACT_BROKER_BASIC_AUTH_PASSWORD)
-          export PACT_PROVIDER_URL=https://$APIGEE_ENVIRONMENT.api.service.nhs.uk/$SERVICE_BASE_PATH
+          export PACT_BROKER_NEXT_URL=$(PACT_BROKER_NEXT_URL)
```
mfjarvis, line 62: Where are these used?

mfjarvis: By which I mean `PACT_BROKER_NEXT_URL` and `PACT_BROKER_NEXT_TOKEN`

tests/e2e/pact/broker/publish.ts
```diff
@@ -20,5 +22,5 @@ async function publish(): Promise<Array<string>> {
 }
 
 (async () => {
```
mfjarvis, line 24: Do we need a final catch to log any errors? Would we get anything useful from that?

azure/azure-release-template.yml
mfjarvis: Ah, I think I've now seen the answer for `PACT_PROVIDER_URL`, but the two `PACT_BROKER_NEXT` ones appear to be unused

tests/e2e/pact/broker/verify.ts
```diff
@@ -1,43 +1,86 @@
...
-  return await verifier.verifyProvider()
+async function verifyProcess(): Promise<any> { 
+  endpoint = "process"
+  sleepMs = 0
+  await verify().catch(verify).catch(verify)
 }
 
-(async () => {
-  verify().catch(verify).catch(verify)
+(async () => {
```
mfjarvis, line 79: Similar thing here, would it be useful to add a catch for process, and add some logging to each catch? Or would that extra logging not add anything?

tests/e2e/pact/specs/sandbox/eps-sandbox.prepare.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+            },
+            method: "POST",
+            path: "/$prepare",
+            body: JSON.parse(requestStr)
+          },
+          willRespondWith: {
+            headers: {
+              "Content-Type": "application/json"
+            },
+            body: {
```
mfjarvis, line 50: What's the reason for this matching the body in a different way to the non-sandbox prepare tests?

coordinator/src/routes/prescription/post-prepare-message.ts
```diff
@@ -13,9 +14,11 @@ export default [
     method: "POST",
     path: "/$prepare",
     handler: validatingHandler(
-      (requestPayload: Bundle, _: Hapi.Request, responseToolkit: Hapi.ResponseToolkit) => {
+      (requestPayload: Bundle, request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit) => {
+        const isSmokeTest = request.headers["x-smoke-test"]
+        const contentType = isSmokeTest ? CONTENT_TYPE_JSON : CONTENT_TYPE_FHIR
```
mfjarvis, line 19: Why did we need to add this? I'm sure there's a good reason, just curious

azure/azure-release-template.yml
booshibooshi: I added these to be able to test out using a different broker to diagnose stability issues without breaking other prs depending on the current broker, but can clean up if you'd rather we didn't have placeholder variables, these are also referenced in publish and verify

mfjarvis: If it's going to be useful in future, leave it in. I just thought it might be vestigial code

coordinator/src/routes/prescription/post-prepare-message.ts
booshibooshi: In order to use the same pact verifier for all pact tests had to add a workaround for the content-type detection, as the latest pact verififer was unable to work out the content-type "application/fhir+json" so the diffs weren't working and pact would fail body matching for everything. Hopefully they'll fix in future version

tests/e2e/pact/broker/publish.ts
booshibooshi: I haven't yet come across publishing pacts failing 3 times in a row, the retry here is for dealing with instability which I think comes with hosting the pact broker in heroku

tests/e2e/pact/broker/verify.ts
booshibooshi: Logging will still come from the pact binary, have found catching an error and logging it here just duplicates whatevers already been pushed to stdout from the pact binary

coordinator/src/routes/prescription/post-prepare-message.ts
mfjarvis: Ah fair enough

mfjarvis (approved):

tests/e2e/pact/broker/verify.ts
mfjarvis: :+1:

tests/e2e/pact/broker/publish.ts
mfjarvis: :+1:

tests/e2e/pact/specs/sandbox/eps-sandbox.prepare.spec.ts
booshibooshi: There was an issue here before which this circumvented but cannot remember what that was now, I'll update this to match live prepare tests and re-run

mfjarvis (approved):

PR #281, opened 2020-12-11
fix url bug and refactor lil bits around

dan-jones-iw:


coordinator/tests/services/translation/cancellation-response/common.spec.ts
```diff
@@ -11,3 +11,17 @@ describe("convertNameUse", () => {
     expect(actual).toEqual([{"family": "last", "given": ["given"], "prefix": ["prefix"]}])
   })
 })
+
+describe("resourceId", () => {
+  test("generate", () => {
+    const resourceId = generateResourceId()
+    expect(resourceId).toHaveLength(36)
+    expect(resourceId).toContain("-")
```
conner-jowitt, line 19: Is it worth testing against regex?

conner-jowitt (approved):

PR #284, opened 2020-12-11
AEA-909

harryhmc:


tests/e2e/pact/specs/eps-convert.spec.ts
```diff
@@ -56,6 +59,41 @@ jestpact.pactWith(
...
+
+      test.each(errorCases)("should receive expected error code in response to %s message", async (desc: string, request: Bundle, response: string, responseMatcher: string, statusCode: string) => {
+        const regex = new RegExp(responseMatcher)
+        const isMatch = regex.test(response)
+        expect(isMatch).toBe(true)
+
+        const requestStr = LosslessJson.stringify(request)
+        const requestJson = JSON.parse(requestStr)
+
+        const interaction = {
```
dan-jones-iw, line 71: reckon we can remove most of the duplication over these interaction objects?

ri-ch (commented):
Can we split this PR up so that it either contains only the test prescriptions, or only the code changes?

mfjarvis (approved):

PR #292, opened 2020-12-16
Adding SSM parameter for ref key and cert

mfjarvis:


mfjarvis:
Laurence confirmed that the int certificate is the correct one to use for ref, so no further changes required

booshibooshi (approved):

PR #307, opened 2020-12-21
AEA-714: Update prepare tests and python script.

mfjarvis:


coordinator/src/services/translation/cancellation/cancellation-medication-request.ts
```diff
@@ -37,6 +37,7 @@ export function createMedicationRequest(
     intent: "order",
     medicationCodeableConcept: getMedicationCodeableConcept(),
     subject: createReference(patientReference),
+    //TODO - effectiveTime should probably be the timestamp of the status, not authoredOn
```
ri-ch, line 40: Shall we create a ticket for this todo?

ri-ch (approved):

PR #308, opened 2020-12-22
Include validator in status check. Make build wait for correct endpoint

ri-ch:


azure/azure-release-template.yml
```diff
@@ -94,7 +94,7 @@ extends:
           make publish-pacts > /dev/null
 
           # Poll deploying API until our deployed version matches the release version
-          url="https://$APIGEE_ENVIRONMENT.api.service.nhs.uk/$SERVICE_BASE_PATH/_ping"
+          url="https://$APIGEE_ENVIRONMENT.api.service.nhs.uk/$SERVICE_BASE_PATH/_status"
```
mfjarvis, line 97: The _status endpoint doesn't include the deployed commit, so I don't think this will work

mfjarvis: Think we need to poll _ping until we know the proxy deploy has happened, and then poll _status until the service has come up

mfjarvis: Ignore me I like jumping to conclusions

mfjarvis (dismissed):

mfjarvis (dismissed):
Reapproving because Rich loves force pushing

azure/azure-release-template.yml
```diff
@@ -44,15 +44,19 @@ extends:
           releaseType: stable
         condition: and(succeeded(), eq(variables['run_smoke_tests'], 'true'))
       - bash: |
+          #!/bin/bash
+
+          # shellcheck disable=SC2155
```
mfjarvis, line 49: Could we not just do what it's telling us to and split the declaration from the assignment?

azure/azure-release-template.yml
```diff
@@ -65,62 +69,74 @@ extends:
           export PACT_PROVIDER_URL=https://$APIGEE_ENVIRONMENT.api.service.nhs.uk/$SERVICE_BASE_PATH
 
           if [[ "$APIGEE_ENVIRONMENT" == *"sandbox"* ]]; then
-            export PACT_PROVIDER=nhsd-apim-eps-sandbox
+          export PACT_PROVIDER=nhsd-apim-eps-sandbox
```
mfjarvis, line 72: We've got a weird mix of indenting here. E.g. line 72 isn't indented but line 77 is.

mfjarvis: I think the indenting makes it easier to keep track of what's going on so my preference would be to keep it

azure/azure-release-template.yml
```diff
@@ -65,62 +69,74 @@ extends:
...
               export IDP_URL="https://nhsd-apim-testing-$APIGEE_ENVIRONMENT.herokuapp.com"
               docker pull artronics/nhsd-login-docker:latest > /dev/null
               export APIGEE_ACCESS_TOKEN=$(docker run --rm artronics/nhsd-login-docker:latest "$IDP_URL")
-            fi
+          fi
           fi
 
           # Publish
-          cd $SERVICE_NAME/$SERVICE_ARTIFACT_NAME/pact
+          cd "$SERVICE_NAME"/"$SERVICE_ARTIFACT_NAME"/pact
```
mfjarvis, line 94: Not a big deal by any means, but we could just enclose the entire thing in double quotes instead of having two sets

azure/azure-release-template.yml
```diff
@@ -65,62 +69,74 @@ extends:
...
-          while true;
+          printf "\nPolling %s every %s seconds, until commit is: %s\n" "$url" "$interval_in_seconds" "$releaseCommit"
+          attempts=0
+          success=0
+          until [ $attempts -eq 15 ]
           do
-            deployedCommit=`curl $url -s | jq -r $path`
-            if [[ "$deployedCommit" == "$releaseCommit" ]]; then
+          responseData=$(curl -H "apiKey: $(status-endpoint-api-key)" "$url" -s)
+          deployedCommit=$(echo $responseData | jq -r ".commitId")
```
mfjarvis, line 110: What happens if the coordinator isn't up yet and so `responseData` is empty?

azure/azure-release-template.yml
```diff
@@ -65,62 +69,74 @@ extends:
...
-          docker run --rm -e PACT_BROKER_BASE_URL=$PACT_BROKER_URL -e PACT_BROKER_USERNAME=$PACT_BROKER_BASIC_AUTH_USERNAME -e PACT_BROKER_PASSWORD=$PACT_BROKER_BASIC_AUTH_PASSWORD pactfoundation/pact-cli:latest broker create-version-tag --pacticipant=$PACT_PROVIDER+prepare+$PACT_VERSION --version=$PACT_VERSION --tag=$APIGEE_ENVIRONMENT
-          docker run --rm -e PACT_BROKER_BASE_URL=$PACT_BROKER_URL -e PACT_BROKER_USERNAME=$PACT_BROKER_BASIC_AUTH_USERNAME -e PACT_BROKER_PASSWORD=$PACT_BROKER_BASIC_AUTH_PASSWORD pactfoundation/pact-cli:latest broker create-version-tag --pacticipant=$PACT_PROVIDER+process+$PACT_VERSION --version=$PACT_VERSION --tag=$APIGEE_ENVIRONMENT
+          docker run --rm -e PACT_BROKER_BASE_URL="$PACT_BROKER_URL" -e PACT_BROKER_USERNAME="$PACT_BROKER_BASIC_AUTH_USERNAME" -e PACT_BROKER_PASSWORD="$PACT_BROKER_BASIC_AUTH_PASSWORD" pactfoundation/pact-cli:latest broker create-version-tag --pacticipant=$PACT_PROVIDER+convert+"$PACT_VERSION" --version="$PACT_VERSION" --tag="$APIGEE_ENVIRONMENT"
```
mfjarvis, line 132: I'd put the double quotes around the entirety of `$PACT_PROVIDER+convert+$PACT_VERSION`. Similarly on the other `docker run` lines

proxies/live/apiproxy/policies/KeyValueMapOperations.GetSecureVariables.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<KeyValueMapOperations async="false" continueOnError="false" enabled="true" name="KeyValueMapOperations.GetSecureVariables" mapIdentifier="pds-variables-encrypted">
```
mfjarvis, line 2: Is this policy supposed to be here, and should we really be using PDS variables?

proxies/live/apiproxy/policies/KeyValueMapOperations.GetVariables.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<KeyValueMapOperations async="false" continueOnError="false" enabled="true" name="KeyValueMapOperations.GetVariables" mapIdentifier="pds-variables">
```
mfjarvis, line 2: Is this policy supposed to be here, and should we really be using PDS variables?

proxies/live/apiproxy/targets/apim.xml
```diff
@@ -1,31 +1,48 @@
...
-      <Step>
-        <Name>AssignMessage.AddIssuerHeader</Name>
-      </Step>
-    </Request>
-  </PreFlow>
+  <Flows>
+    <Flow name="Status">
+      <Request>
+        <Step>
+          <Name>KeyValueMapOperations.GetSecureVariables</Name>
```
mfjarvis, line 6: Again, don't think we're actually using this

azure/azure-release-template.yml
ri-ch: We could. I googled the error and it said it was a false positive a lot of the time, but i suppose we may aswell.

azure/azure-release-template.yml
ri-ch: Not 'nicely' really. Curl returns a load of crap and then JQ fails to extract the expected fields

mfjarvis (approved):

PR #311, opened 2020-12-30
AEA-401: Validate MedicationRequest status and intent

mfjarvis:


coordinator/tests/services/validation/bundle-validator.spec.ts
```diff
@@ -44,7 +48,69 @@ describe("Bundle checks", () => {
   })
 })
 
-describe("MedicationRequest checks", () => {
+describe("verifyCommonBundle", () => {
```
dan-jones-iw, line 51: `"verifyCommonBundle order check"`

dan-jones-iw (approved):

PR #312, opened 2020-12-31
AEA-924: Return HealthcareService instead of Organization in cancel responses.

mfjarvis:


coordinator/src/services/translation/cancellation/cancellation-organization.ts
```diff
@@ -1,7 +1,7 @@
 import * as fhir from "../../../models/fhir/fhir-resources"
 import * as hl7 from "../../../models/hl7-v3/hl7-v3-people-places"
 import {convertAddress, convertTelecom, generateResourceId} from "./common"
-import {createIdentifier} from "./fhir-base-types"
+import {createIdentifier, createReference} from "./fhir-base-types"
 
 export function createOrganization(hl7Organization: hl7.Organization): fhir.Organization {
```
dan-jones-iw, line 6: is the createOrganization function gonna be used anymore?

dan-jones-iw (approved):

coordinator/src/services/translation/cancellation/cancellation-organization.ts
mfjarvis: Probably not any time soon, but it will be if we start supporting primary care

PR #314, opened 2021-01-04
aea - remove bundle entry duplication

dan-jones-iw:
remove duplication of certain resources in fhir response from spine and update tests to reflect changes

### DIFF
--- a/coordinator/src/services/translation/cancellation/cancellation-response.ts
+++ b/coordinator/src/services/translation/cancellation/cancellation-response.ts
@@ -44,12 +44,15 @@ function convertResourceToBundleEntry(resource: fhir.Resource) {
 function createBundleEntries(cancellationResponse: CancellationResponse) {
   const fhirPatient = createPatient(cancellationResponse.recordTarget.Patient)
 
+  const hl7ResponsiblePartyAgentPerson = cancellationResponse.responsibleParty.AgentPerson
+  const hl7AuthorAgentPerson = cancellationResponse.author.AgentPerson
+
   const {
     fhirPractitioner: fhirResponsiblePartyPractitioner,
     fhirLocations: fhirResponsiblePartyLocations,
     fhirHealthcareService: fhirResponsiblePartyHealthcareService,
     fhirPractitionerRole: fhirResponsiblePartyPractitionerRole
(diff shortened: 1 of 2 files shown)


coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -76,39 +100,40 @@ function createBundleEntries(cancellationResponse: CancellationResponse) {
...
-    fhirAuthorPractitioner,
-    ...fhirAuthorLocations,
-    fhirAuthorHealthcareService,
-    fhirAuthorPractitionerRole,
-    fhirResponsiblePartyPractitioner,
-    ...fhirResponsiblePartyLocations,
-    fhirResponsiblePartyHealthcareService,
-    fhirResponsiblePartyPractitionerRole
-  ]
+  bundleResources.push(fhirMedicationRequest, fhirMessageHeader)
```
conner-jowitt, line 103: I think the message header needs to be the first resource in the bundle: https://www.hl7.org/fhir/messageheader.html "The reference(s) that are the subject of the action as well as other information related to the action are typically transmitted in a bundle in which the MessageHeader resource instance is the first resource in the bundle."

coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -50,18 +51,41 @@ function createBundleEntries(cancellationResponse: CancellationResponse) {
...
+    fhirResponsiblePartyPractitioner,
+    ...fhirResponsiblePartyLocations,
+    fhirResponsiblePartyHealthcareService,
+    fhirResponsiblePartyPractitionerRole
+  ]
+
+  let requesterId = fhirResponsiblePartyPractitioner.id
+
+  if (!isDeepStrictEqual(
+    cancellationResponse.responsibleParty.AgentPerson,
```
conner-jowitt, line 65: Would this look neater if `cancellationResponse.responsibleParty.AgentPerson` and `cancellationResponse.author.AgentPerson` were pulled out into a variable? They could then be used in the performer checks too.

conner-jowitt (approved):

conner-jowitt (approved):

PR #315, opened 2021-01-04
Aea idk push multiple prescriptions to int

dan-jones-iw:


coordinator/src/models/fhir/fhir-resources.ts
```diff
@@ -78,10 +82,11 @@ export interface IdentifierReference<T extends Resource> {
 export interface Dosage {
   text: string
   patientInstruction?: string
+  additionalInstruction?: Array<CodeableConcept>
 }
 
 export interface Performer extends IdentifierReference<Organization> {
-  extension: Array<ReferenceExtension<PractitionerRole>>
+  extension?: Array<ReferenceExtension<PractitionerRole>>
```
conner-jowitt, line 89: Does it make sense that the only field in this interface is nullable?

dan-jones-iw: had to make this change since i was building medicationRequest objects that didnt have a performer block in them (due to difference between prescription-order & prescription-order-update MRs). i guess it makes sense in a way since a performer block is just a particular IdentifierReference that may have an extension block too.

conner-jowitt: Think I understand. It might be clearer if this was renames `IdentifierReferenceWithExtension` (or something similar) and the extension wasn't nullable. Then `Performer` could be `IdentifierReferenceWithExtension || Identifierreference`. Not sure, but that feels neater to me?

conner-jowitt: Maybe I'm just overthinking it

conner-jowitt: and it might confuse things further down the line, idk

coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -48,43 +48,43 @@ function createBundleEntries(cancellationResponse: CancellationResponse) {
...
   const {
-    fhirPractitioner: fhirResponsiblePartyPractitioner,
-    fhirLocations: fhirResponsiblePartyLocations,
-    fhirHealthcareService: fhirResponsiblePartyHealthcareService,
-    fhirPractitionerRole: fhirResponsiblePartyPractitionerRole
+    fhirPractitioner: fhirCancelRequesterPractitioner,
+    fhirLocations: fhirCancelRequesterLocations,
+    fhirHealthcareService: fhirCancelRequesterHealthcareService,
+    fhirPractitionerRole: fhirCancelRequesterPractitionerRole
   } = convertAgentPerson(hl7ResponsiblePartyAgentPerson)
```
mfjarvis, line 55: This is the wrong way round. ResponsibleParty in the HL7 V3 message is the author of the original prescription. Author is the author of the cancellation message.

coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -48,43 +48,43 @@ function createBundleEntries(cancellationResponse: CancellationResponse) {
...
     const {
-      fhirPractitioner: fhirAuthorPractitioner,
-      fhirLocations: fhirAuthorLocations,
-      fhirHealthcareService: fhirAuthorHealthcareService,
-      fhirPractitionerRole: fhirAuthorPractitionerRole
+      fhirPractitioner: fhirOriginalPrescriptionAuthorPractitioner,
+      fhirLocations: fhirOriginalPrescriptionAuthorLocations,
+      fhirHealthcareService: fhirOriginalPrescriptionAuthorHealthcareService,
+      fhirPractitionerRole: fhirOriginalPrescriptionAuthorPractitionerRole
     } = convertAgentPerson(hl7AuthorAgentPerson)
```
mfjarvis, line 73: As above

coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -44,47 +44,47 @@ function convertResourceToBundleEntry(resource: fhir.Resource) {
...
-    fhirResponsiblePartyHealthcareService,
-    fhirResponsiblePartyPractitionerRole
+    fhirCancelRequesterPractitioner,
+    ...fhirCancelRequesterLocations,
+    fhirCancelRequesterHealthcareService,
+    fhirCancelRequesterPractitionerRole
   ]
 
-  let requesterId = fhirResponsiblePartyPractitioner.id
+  let requesterId = fhirCancelRequesterPractitioner.id
```
mfjarvis, line 65: maybe rename to authorId? (sorry)

coordinator/src/services/translation/cancellation/cancellation-response.ts
```diff
@@ -110,7 +110,7 @@ function createBundleEntries(cancellationResponse: CancellationResponse) {
     const performerAgentPerson = cancellationResponse.performer.AgentPerson
     let performerId
     if (isDeepStrictEqual(performerAgentPerson, hl7ResponsiblePartyAgentPerson)) {
```
mfjarvis, line 112: I think this should be checking hl7AuthorAgentPerson on line 112 and hl7ResponsiblePartyAgentPerson on line 114

coordinator/src/services/translation/cancellation/cancellation-response.ts
dan-jones-iw: yea i was rereading the code and requester is pretty confusing here - its requesterId at the moment because its used to populate the `requester` field on a MedicationRequest

mfjarvis (approved):

PR #318, opened 2021-01-06
remove the spine message from a process-message response

dan-jones-iw:


coordinator/src/services/translation/spine-response.ts
```diff
@@ -86,7 +86,7 @@ function getFhirResponseAndErrorCodes<T extends AsyncMCCI | SyncMCCI>(
 
 function createOperationOutcomeIssue(
   statusCode: number,
-  hl7Message: string,
+  hl7Message?: string,
```
mfjarvis, line 89: Is there a reason we're still passing this in on line 76 & 77? I'd have thought we'd want to get rid of it in all places

dan-jones-iw: i was only getting rid of it on successful prescription-order, but on second thoughts i'll remove it entirely

mfjarvis (approved):

PR #323, opened 2021-01-11
added endpoint for container metrics

conner-jowitt:


coordinator/src/routes/health/get-metrics.ts
```diff
@@ -0,0 +1,26 @@
+import Hapi from "@hapi/hapi"
+import * as osu from "node-os-utils"
+
+export default [
+  {
+    method: "GET",
+    path: "/_status",
```
dan-jones-iw, line 7: should this have a different path assuming that the status route still exists?

conner-jowitt: Yep, updated

dan-jones-iw (dismissed):

coordinator/src/routes/health/get-metrics.ts
```diff
@@ -0,0 +1,26 @@
+import Hapi from "@hapi/hapi"
+import * as osu from "node-os-utils"
+
+export default [
+  {
+    method: "GET",
+    path: "/_metrics",
+    handler: async (request: Hapi.Request, h: Hapi.ResponseToolkit): Promise<Hapi.ResponseObject> => {
+      const cpu = osu.cpu
```
ri-ch, line 9: Not that it matters (the compiler will probably take out the interim field for you) but is there any reason we store a reference to `cpu` then call info on that instead of just doing `osu.cpu.info()`?

ri-ch (approved):

ri-ch (changes requested):
The only thing I would say is that the contents of the metrics endpoint are probably pretty sensitive, so we probably want to stop people calling it from the internet (at least with out an access token). We could copy the same method we have for the _status endpoint?

coordinator/src/routes/health/get-status.ts
```diff
@@ -1,28 +1,30 @@
...
+  try {
+    request.logger.info("Checking validator status")
+    const response = await axios.get<string>(`${VALIDATOR_HOST}/_status`, {timeout: 2})
+
+    if (response.status == 200 && response.data == "Validator is alive") {
+      return true
+    } else {
+      const responseSummary = `Status: ${response.status}, data: ${response.data ?? "No Data"}`
+      const msg = `Did not get positive response from validator status check. ${responseSummary}`
+      request.logger.warn(msg)
```
dan-jones-iw, line 14: what gets returned here?

ri-ch (approved):

ri-ch (commented):
Still need the change to proxy

ri-ch (changes requested):
Update proxy to secure metrics enpoint

ri-ch (approved):

PR #325, opened 2021-01-11
Add auditing for inbound and outbound messages

ri-ch:


coordinator/src/routes/prescription/post-process-message.ts
```diff
@@ -15,6 +15,8 @@ export default [
       async (requestPayload: Bundle, request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit) => {
         const spineRequest = translator.convertFhirMessageToSpineRequest(requestPayload)
         const spineResponse = await requestHandler.send(spineRequest, request.logger)
+        request.log("audit", {messageType: "Inbound FHIR message", payload: requestPayload})
```
dan-jones-iw, line 18: wont this contain potentially sensitive information?

dan-jones-iw: same as with the other logs

ri-ch: Yea, thats the only issue with this approach. It _should_ be ok, given the only people with access to production logs are security cleared

dan-jones-iw: happy to approve if we are sure people with access to production logs are security cleared

coordinator/src/routes/prescription/post-prepare-message.ts
```diff
@@ -18,6 +18,8 @@ export default [
         const isSmokeTest = request.headers["x-smoke-test"]
         const contentType = isSmokeTest ? CONTENT_TYPE_JSON : CONTENT_TYPE_FHIR
         const response = translator.convertFhirMessageToSignedInfoMessage(requestPayload)
+        request.log("audit", {messageType: "Inbound FHIR message", payload: requestPayload})
+        request.log("audit", {messageType: "Outbound signed info message", outbound: response})
```
harryhmc, line 22: am i right in thinking this will make a call to hapi's own logger and not the pino instance / does that matter?

ri-ch: It goes through Hapi, but pino acts as the default logger anyway

coordinator/src/routes/prescription/post-process-message.ts
conner-jowitt: There was a discussion on Slack on 06/01 about user data in logs, think we decided it all had to be removed? We've merged in a PR to remove  messages from other log events

PR #331, opened 2021-01-18
explain signing service integration and generation of XML-DSig

dan-jones-iw:


specification/electronic-prescription-service-api.yaml
```diff
@@ -37,14 +37,25 @@ info:
...
-    4. POST the signed prescription-order message to the `/$process-message` endpoint
+    1. POST a [prescription-order message](https://simplifier.net/guide/DigitalMedicines/prescription-order) to the `/$prepare` endpoint to get the encoded prescription data that requires a signature
+    2. sign the encoded prescription data using the Signing Service to get the signed prescription data
+    3. generate an XML-DSig using the signed prescription data (see below)
+    4. put the XML-DSig in a `Provenance` resource
+    5. add the `Provenance` to the prescription-order message
+    6. POST the signed prescription-order message to the `/$process-message` endpoint
+
+    Example code to generate an XML-DSig: 
+    ```private fun toXmlSignature(digest: ByteArray, signature: ByteArray, certificate: ByteArray): ByteArray {
```
mfjarvis, line 52: I like the code example :+1:.

Would be good if we told the supplier where to get each of the three parameters. E.g. "digest" comes from EPS `POST /$prepare`, "signature" and "certificate" come from Signing Service `GET SignatureResponse`.

Also, the method could probably take three strings instead of three ByteArrays. The "digest" which comes back from prepare is already a string, and the "signature" and "certificate" which come back from SignatureResponse are already base64 encoded strings.

mfjarvis: Also we should base64 encode the final result before it goes in the Provenance.signature.data.

tjhmoore (approved):

specification/electronic-prescription-service-api.yaml
```diff
@@ -37,14 +37,31 @@ info:
...
+
+    Example code to generate an XML-DSig: 
+    ```javascript
+      function toXmlSignature(digest: string, signature: string, certificate: string): string {
+        const decodedDigest = Buffer.from(digest, "base64").toString("utf-8") 
+        const digestWithoutNamespace = decodedDigest.replace("<SignedInfo xmlns=\"http://www.w3.org/2000/09/xmldsig#\">", "<SignedInfo>")
+        const xmlSignature = `<Signature xmlns="http://www.w3.org/2000/09/xmldsig#">${digestWithoutNamespace}<SignatureValue>${signature}</SignatureValue><KeyInfo><X509Data><X509Certificate>${certificate}</X509Certificate></X509Data></KeyInfo></Signature>`
+        return Buffer.from(xmlSignature, "utf-8").toString("base64")
+      }
+    ```
```
SachinPath, line 64: Point 3 within the 'To generate an XML-DSig:', needs to be changed from (javascript example seen below) to (javascript example see below).

mick-sch (approved):
Slightly odd step ordering but I see why you did it.

PR #333, opened 2021-01-20
Improve testing tooling

booshibooshi:


make.ps1
```diff
@@ -0,0 +1,117 @@
+# Run the command below to add make facade commands to powershell:
```
ri-ch, line 1: Isnt there a native make binary for windows?

booshibooshi: as discussed, thanks for pointing me to https://chocolatey.org/packages/make. This definitely seems to be the way to go to keep things simple but will require re-writing Makefiles with cross-platform in mind

ri-ch (approved):
Approved in the interests of pragmatism, but we need to go back and revisit these changes

PR #335, opened 2021-01-21
Expose validator metrics

ri-ch:


coordinator/src/routes/health/get-status.ts
```diff
@@ -30,5 +30,25 @@ export default [
         commitId: process.env.COMMIT_ID
       })
     }
+  },
+  {
+    method: "GET",
+    path: "/_validatormetrics/{path*}",
```
dan-jones-iw, line 36: `/_validatormetrics` seems a bit ugly, can we capitalise M maybe? or use a dash like in `/$process-message`?

proxies/live/apiproxy/targets/apim.xml
```diff
@@ -10,7 +10,7 @@
           <Name>RaiseFault.403Forbidden</Name>
         </Step>
       </Request>
-      <Condition>(proxy.pathsuffix MatchesPath "/_status") and ((request.verb = "GET") or (request.verb = "HEAD"))</Condition>
+      <Condition>(proxy.pathsuffix MatchesPath "/_status") and (proxy.pathsuffix MatchesPath "/_validatormetrics") and ((request.verb = "GET") or (request.verb = "HEAD"))</Condition>
```
dan-jones-iw, line 13: how can `proxy.pathsuffix` match both `/_status` and `/_validatormetrics`

coordinator/src/routes/health/get-status.ts
dan-jones-iw: its not super clear what kind of values `path` would have? are there multiple endpoints on the validator that report metrics?

dan-jones-iw (dismissed):

proxies/live/apiproxy/targets/apim.xml
ri-ch: Good spot. It cant. That should be an 'or' in between.

coordinator/src/routes/health/get-status.ts
ri-ch: It is a bit ugly. I dont think the name is too important, this is an internal api for diagnostics, and we might not even go to production with it

ri-ch: `{path*}` just picks up any path components and proxies them onto the metrics endpoint

dan-jones-iw (approved):

PR #340, opened 2021-01-28
AEA-1045: Return 500 from _status endpoint if service is not healthy.

mfjarvis:


coordinator/src/routes/health/get-status.ts
```diff
@@ -11,8 +11,7 @@ export default [
 
       try {
         request.logger.info("Checking validator status")
-        const response = await axios.get<string>(`${VALIDATOR_HOST}/_status`, {timeout: 2})
-
+        const response = await axios.get<string>(`${VALIDATOR_HOST}/_status`, {timeout: 20000})
```
conner-jowitt, line 14: Does this timeout need to be so long? Is it just to stop AWS polling the status endpoint as often during setup?

mfjarvis: I just used the same value PDS used for their status endpoint. No particular reason for 20s. Just wanted something less than 30s because that's when clients might start to time out waiting for the response

mfjarvis: Also don't want to go too low because when the container's under stress the response time will be high, but we don't want ECS to kill it as a result

conner-jowitt (approved):

PR #341, opened 2021-01-29
Aea 550 add X-Request-ID header

conner-jowitt:


mfjarvis:
Looks good. I'd consider: 
- making X-Request-ID mandatory at the proxy level so we also get one for prepare and poll requests
- echoing the X-Request-ID header in the response, again in the proxy (PDS does this)
but I'm not sure those need to be done as part of this PR.

proxies/live/apiproxy/targets/apim.xml
```diff
@@ -21,6 +21,12 @@
           <Name>RaiseFault.403Forbidden</Name>
           <Condition>!(accesstoken.id_token-acr StartsWith "AAL3")</Condition>
         </Step>
+        <Step>
+          <!-- Header NHSD-Request-ID must be in the correct format (GUID) -->
+          <Name>RaiseFault.400BadRequest</Name>
+          <Condition>(original-request-details.header.X-Request-ID != null) and (not original-request-details.header.X-Request-ID ~~ "^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")</Condition>
```
mfjarvis, line 27: I think we're missing the step which creates `original-request-details`. PDS has a policy named `AssignMessage.Copy.RequestHeaders` which does this bit.

tests/e2e/pact/specs/live/eps.convert.spec.ts
```diff
@@ -43,14 +46,16 @@ jestpact.pactWith(
...
 
         const apiPath = "/$convert"
         const interaction: InteractionObject = {
           state: "is authenticated",
           uponReceiving: `a request to convert ${desc} message`,
           withRequest: {
             headers: {
-              "Content-Type": "application/fhir+json; fhirVersion=4.0"
+              "Content-Type": "application/fhir+json; fhirVersion=4.0",
+              "X-Request-ID": requestId
```
mfjarvis, line 58: Maybe add `X-Request-ID` to the `willRespondWith` block as well to check it's present in the response. Could also test `X-Correlation-ID` in a similar way.

proxies/live/apiproxy/targets/apim.xml
mfjarvis: Although the smoke tests didn't fail so maybe I missed something

coordinator/src/services/handlers/spine-handler.ts
```diff
@@ -22,20 +22,20 @@ const httpsAgent = new https.Agent({
 export class LiveRequestHandler implements RequestHandler {
   private readonly spineEndpoint: string
   private readonly spinePath: string
-  private readonly ebXMLBuilder: (spineRequest: SpineRequest) => string
+  private readonly ebXMLBuilder: (spineRequest: SpineRequest, xRequestIdHeader: string) => string
```
ri-ch, line 25: At this point, its not a header any more, its just a plain request ID. Could we rename this to `requestId`?

coordinator/src/services/handlers/spine-handler.ts
```diff
@@ -22,20 +22,20 @@ const httpsAgent = new https.Agent({
...
 
   constructor(
     spineEndpoint: string = null,
     spinePath: string = null,
-    ebXMLBuilder: (spineRequest: SpineRequest) => string = null
+    ebXMLBuilder: (spineRequest: SpineRequest, xRequestIdHeader: string) => string = null
```
ri-ch, line 30: Could we update the definition of spine request to also include a field for spine request so that we don't need to independently pass around the request ID?

coordinator/src/services/translation/prescription/cancellation.ts
```diff
@@ -16,8 +16,14 @@ export function convertCancellation(
   const fhirFirstMedicationRequest = getMedicationRequests(fhirBundle)[0]
   const effectiveTime = extractEffectiveTime(fhirFirstMedicationRequest)
 
+  const messageId = getIdentifierValueForSystem(
```
ri-ch, line 19: Seems like we have to do this twice, is there any way we can centralise this logic?

tests/e2e/pact/specs/live/eps.convert.spec.ts
```diff
@@ -76,22 +83,25 @@ jestpact.pactWith(
 
         const requestStr = LosslessJson.stringify(request)
         const requestJson = JSON.parse(requestStr)
+        const requestId = uuid.v4().toString().toLowerCase()
```
ri-ch, line 86: We do this a lot, can we have a `uuidstring` function somewhere so we dont need to keep typing this.

mfjarvis: `uuid.v4()` already gives you a string and a lowercase one at that

tests/e2e/pact/specs/live/eps.convert.spec.ts
```diff
@@ -26,10 +27,12 @@ jestpact.pactWith(
       test(authenticationTestDescription, async () => {
         const apiPath = "/$convert"
         const interaction: InteractionObject = createUnauthorisedInteraction(authenticationTestDescription, apiPath)
+        const requestId = uuid.v4().toString().toLowerCase()
         await provider.addInteraction(interaction)
         await client()
           .post(apiPath)
           .set('Content-Type', 'application/fhir+json; fhirVersion=4.0')
+          .set('X-Request-ID', requestId)
```
booshibooshi, line 35: could move this to requestFilter so it's only set in one place, see verify.ts

booshibooshi: unless we're checking the response guid too as suggested by Matt below, in which case here is fine

conner-jowitt: We are checking the X-Request-ID header matches in the request and response, I'll add in the check for X-Correlation-ID too. Is there a nice way to centralise doing those checks or would it require a lot of refactoring?

tests/e2e/pact/specs/live/eps.convert.spec.ts
conner-jowitt: The function you're thinking of is in the coordinator which gets packaged separately from the PACTs, I can duplicate the function if we think it's needed. I'll remove all the `.toString().toLowerCase()` if it already does that, I just pulled it from the stuff in coordinator

tests/e2e/pact/specs/live/eps.convert.spec.ts
booshibooshi: Could look at wrapping that client in a function inside the common.ts Dan's just added

coordinator/src/services/translation/prescription/cancellation.ts
conner-jowitt: Yep, changed

tests/e2e/pact/specs/live/eps.convert.spec.ts
conner-jowitt: Added

coordinator/src/services/handlers/spine-handler.ts
conner-jowitt: Done

booshibooshi (approved):

PR #343, opened 2021-02-02
AEA-949: New UK Core version. Removing schema from OpenAPI spec.

mfjarvis:


pyproject.toml
```diff
@@ -1,25 +1,20 @@
...
-
 [tool.poetry]
-name = "apim-definitions"
+name = "electronic-prescription-service-api"
 version = "0.0.1-alpha"
 description = "Electronic Prescription Service API Definitions"
-
 license = "MIT"
-
 authors = [
```
dan-jones-iw, line 9: alphabetical ordering or order of importance??

mfjarvis: Same ordering as https://python-poetry.org/docs/pyproject/, so I kept it

dan-jones-iw (approved):

Makefile
```diff
@@ -150,7 +150,7 @@ export APIGEE_ENVIRONMENT=$(env)
 export APIGEE_ACCESS_TOKEN=$(token)
 
 space :=
-space += 
+space +=
```
mfjarvis, line 153: Dodgy merge conflict resolution?

booshibooshi: Second space should be there and be equal to " ", it's a makefile workaround to remove a space from the username variable, cannot set soace variable to " " itself but you can append a space to an existing blank variable. Sure there's a better way of achieving this

mfjarvis: Nope it's intellij being helpful

dan-jones-iw (approved):

dan-jones-iw (approved):

Makefile
mfjarvis: @alexmccormick-IW github didn't feel like showing me your message. I knew it was supposed to be there, the trouble was stopping intellij from removing it every time we commit :joy: Found another hacky way of defining a space instead

PR #344, opened 2021-02-04
AEA-1059: Add GP acute example.

mfjarvis:
Branched from AEA-949 as this dpeneds on UK Core changes.


models/examples/primary-care/acute/no-nominated-pharmacy/1-Prepare-Request-200_OK.json
```diff
@@ -0,0 +1,526 @@
...
+          "extension": [
+            {
+              "url": "https://fhir.nhs.uk/StructureDefinition/Extension-DM-PerformerSiteType",
+              "valueCoding": {
+                "system": "https://fhir.nhs.uk/CodeSystem/dispensing-site-preference",
+                "code": "P1"
+              }
+            }
+          ],
+          "validityPeriod": {
```
mfjarvis, line 158: Should update this to be realistic (should be later than authoredOn), or remove entirely. Is there any reason for this prescription to be future dated?

booshibooshi: Not sure if I'm answering your question or just adding different context to it but there is validation downstream for start and end, update-prescriptions sets validityPeriod.start to todays date and validityPeriod.end to 4 weeks from today and that's accepted does that sound reasonable enough? 4 weeks from today is set to give a large window for testing dispensing

mfjarvis: Updated to be 4 weeks, starting at authoredOn.

coordinator/src/services/translation/prescription/organization.ts
```diff
@@ -48,10 +56,29 @@ function isNhsTrust(fhirOrganization: fhir.Organization) {
...
-  const fhirParentOrganization = fhirOrganization.partOf
-    ? resolveReference(fhirBundle, fhirOrganization.partOf)
-    : fhirOrganization
+  let fhirParentOrganization = fhirOrganization
+  const partOf = fhirOrganization.partOf
+  if (partOf) {
+    if (isDirectReference(partOf)) {
+      fhirParentOrganization = resolveReference(fhirBundle, partOf)
+    } else {
+      //TODO - Fix error FHIR paths for this case?
```
dan-jones-iw, line 72: does this still need doing?

dan-jones-iw (approved):

coordinator/src/services/translation/prescription/organization.ts
mfjarvis: Would be nice :shrug: Basically if we don't, the errors will come back with a path of e.g. `Organization.identifier` instead of `Organization.partOf.identifier`

mfjarvis: But it was kinda annoying to implement

PR #347, opened 2021-02-05
Aea update spec from rob review

dan-jones-iw:


specification/electronic-prescription-service-api.yaml
```diff
@@ -350,6 +319,8 @@ paths:
         This returns a set of prescriptions based on the information passed to the endpoint: 
         * to acquire prescription information for an entire pharmacy, send owner information
         * to acquire prescription information for a specific prescription, send the group identifier for that prescription
+        
+        This endpoint is only available in sandbox mode.
```
SachinPath, line 323: sandbox mode or sandbox environment?

conner-jowitt (dismissed):

specification/electronic-prescription-service-api.yaml
```diff
@@ -177,62 +178,23 @@ info:
...
-
 
+    ## Errors
+    ### API specific errors
+    | HTTP Status   | Error Code             | Description                                    |
+    | --------------| ---------------------- | ---------------------------------------------- |
+    | 400           | MISSING_FIELD          | A required field is missing within the request |
+    | 400           | INVALID_VALUE          | Invalid value in message                       |
+    | 400           | INCORRECT_RESOURCETYPE | FHIR resource has incorrect resourceType       |
+    | 404           | NOT_FOUND              | Resource not found                             |
```
mfjarvis, line 189: I think `404` can be removed from the API-specific section

dan-jones-iw: this is here since a different error is returned from Apigee vs Hapi 404s, not much value here tho so can remove :+1:

specification/electronic-prescription-service-api.yaml
```diff
@@ -177,62 +178,23 @@ info:
...
+    
+    ### Platform-wide errors
+    
+    | HTTP Status   | Error Code      | Description                                            |
+    | ------------- | --------------- | ------------------------------------------------------ |
+    | 400           | MISSING_FIELD   | X-Request-ID header missing or malformed               |
+    | 401           | ACCESS_DENIED   | User does not have permission for a particular request |
+    | 404           | NOT_FOUND       | Resource not found                                     |
```
mfjarvis, line 197: Could add `403` which happens if a user doesn't use a strong enough authentication method (e.g. username & password instead of smartcard)

dan-jones-iw: nice :+1: would like to try replicate that to see what that looks like

specification/electronic-prescription-service-api.yaml
```diff
@@ -320,7 +282,13 @@ paths:
                   value:
                     $ref: components/examples/secondary-care/community/acute/nominated-pharmacy/nurse/prescribing-and-professional-codes/1-Process-Response-Send-200_OK.json
         '4XX':
-          description: Invalid request.
+          description: |
+            An error occurred as follows:
+            
+            | HTTP status | Error code         | Description                                    | FHIR type        |
+            | ----------- | ------------------ | ---------------------------------------------- | -----------------|
+            | 400         | BAD_REQUEST        | Spine Error - duplicate prescription ID exists | OperationOutcome |
```
mfjarvis, line 290: Don't really understand the `FHIR type` bit of this. What's it telling me?

dan-jones-iw: supposed to explain to a user that some errors will be OperationOutcomes and some Bundles - this hopefully is temporary until we clean up all errors being OperationOutcomes

mfjarvis: Also does this section override the `API specific errors` bit above or add to it? Because we've defined `400` responses in that section

mfjarvis: Ah okay, we could use the examples or schema to show that there are multiple possible error formats, but this is fine for now

mfjarvis (approved):

specification/electronic-prescription-service-api.yaml
dan-jones-iw: yea i agree with this, might merge in and suggest to sachin to have a look into that :+1:

PR #349, opened 2021-02-09
AEA-1046: Add FHIR R4 prefix to proxy route rule.

mfjarvis:


booshibooshi (approved):

mfjarvis:
Rolled into AEA-1120 as this also requires smoke test changes.

PR #350, opened 2021-02-09
AEA-1120: Fix proxy route rule. Change smoke tests to look for success.

mfjarvis:


tests/e2e/pact/specs/live/process.successes.spec.ts
```diff
@@ -44,9 +62,13 @@ processPactGroups.forEach(pactGroup => {
           const requestId = uuid.v4()
           const correlationId = uuid.v4()
 
+          const firstMedicationRequest = message.entry.map(e => e.resource)
```
dan-jones-iw, line 65: would be nice to wrap this into a cute function

dan-jones-iw (approved):
approved for merge - comment to change later

PR #352, opened 2021-02-10
AEA-1121 Add tests to update prescriptions and Add cancelation smoke-tests back

booshibooshi:


Makefile
```diff
@@ -14,8 +14,9 @@ install: install-validator install-node install-python install-hooks
 
 build: build-specification build-coordinator build-validator build-proxies
 
-test: lint validate-models check-licenses test-coordinator
```
booshibooshi, line 17: Removal of lint here as lint calls build and the ADO pipeline calls lint then test so the pipeline is building everything twice. Todo: check commit hook still lints

pyproject.toml
```diff
@@ -1,38 +1,39 @@
-[project]
```
booshibooshi, line 1: Adding pytest dependency re-generated this file

scripts/build-coordinator.ps1
```diff
@@ -1,5 +1,4 @@
 npm run --prefix=coordinator/ build
 cp coordinator/package.json coordinator/dist/
 mkdir -Force coordinator/dist/resources
-cp coordinator/src/resources/ebxml_request.mustache coordinator/dist/resources/
-cp coordinator/src/resources/message_display.mustache coordinator/dist/resources/
```
booshibooshi, line 5: We no longer use display mustache template

scripts/update_prescriptions.py
```diff
@@ -15,16 +15,17 @@
...
 
-def generate_short_form_id():
+def generate_short_form_id(organisationCode):
     """Create R2 (short format) Prescription ID
     Build the prescription ID and add the required checkdigit.
     Checkdigit is selected from the PRESCRIPTION_CHECKDIGIT_VALUES constant
     """
     _PRESCRIPTION_CHECKDIGIT_VALUES = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ+'
     hex_string = str(uuid.uuid1()).replace('-', '').upper()
-    prescription_id = f'{hex_string[:6]}-Z{hex_string[6:11]}-{hex_string[12:17]}'
```
booshibooshi, line 27: Fix short form prescription id missing organisation code in middle third

tests/e2e/pact/broker/verify.ts
```diff
@@ -94,15 +102,17 @@ async function verifyPrepare(): Promise<any> {
...
+          .filter(g => g !== "failures")
+          .filter(g => !g.includes("-cancel"))
       : PactGroups
 
     await pactGroups.reduce(async (promise, group) => {
       await promise
       endpoint = "process"
       pactGroup = group
       resetBackOffRetryTimer()
-      await verifyWith2Retries()
```
booshibooshi, line 105: If process message smoke-test fails the first time it will almost definitely fail the second time as some prescriptions will have been sent and will now be considered "duplicate", or fail a subsequent time as a result of a cached response for the x-request-id in spine. This change gives us a cleaner log to look at when something goes wrong

tests/e2e/pact/specs/live/process.successes.spec.ts
```diff
@@ -29,14 +29,6 @@ const orderUpdatePactGroups = [
...
     name: "secondarycare-community-acute-cancel",
     cases: TestResources.processSecondaryCareCommunityAcuteOrderUpdateCases
-  },
-  {
-    name: "secondarycare-community-repeatdispensing-cancel",
-    cases: TestResources.processSecondaryCareCommunityRepeatDispensingOrderUpdateCases
-  },
-  {
-    name: "secondarycare-homecare-cancel",
-    cases: TestResources.processSecondaryCareHomecareOrderUpdateCases
```
booshibooshi, line 39: These groups have no cancel examples at the moment

Makefile
```diff
@@ -14,8 +14,9 @@ install: install-validator install-node install-python install-hooks
...
+test: validate-models check-licenses test-coordinator
 	cd tests/e2e/pact && make test
+	poetry run pytest -q .\scripts\update_prescriptions.py
```
booshibooshi, line 19: Update-prescriptions tests

Makefile
booshibooshi: checked, hook calls lint directly

PR #354, opened 2021-02-10
Aea release validation update

dan-jones-iw:


coordinator/src/routes/dispention/release.ts
```diff
@@ -29,26 +30,36 @@ export default [
...
+    handler: async (request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit) => {
+      const fhirValidatorResponse = await externalFHIRValidation(request)
+      if (fhirValidatorResponse.issue.length > 0) {
+        return responseToolkit.response(fhirValidatorResponse).code(400)
       }
-    )
+
+      const requestPayload = getPayload(request) as fhir.Resource
+
+      if (requestPayload.resourceType !== "Parameters") {
```
mfjarvis, line 41: Should we do the same thing for `$process-message` to check the message is a `Bundle`?

coordinator/src/models/errors/validation-errors.ts
```diff
@@ -58,10 +58,23 @@ export class MedicationRequestIncorrectValueError implements ValidationError {
...
   operationOutcomeCode = "value" as const
   severity = "fatal" as const
   expression = ["Bundle.entry.resource.ofType(MessageHeader).eventCoding.code"]
 }
+
+export class ResourceTypeError implements ValidationError {
+  message: string
+  operationOutcomeCode = "value" as const
+  severity = "error" as const
+  expression = [""]
```
mfjarvis, line 75: If typescript will let you, I think an empty array would be nicer than an array with a blank entry.

coordinator/src/models/errors/validation-errors.ts
```diff
@@ -58,10 +58,23 @@ export class MedicationRequestIncorrectValueError implements ValidationError {
 export class MessageTypeError implements ValidationError {
   message = `MessageHeader.eventCoding.code must be one of '${
```
ri-ch, line 59: Maybe not for this PR, but would it be nicer to use a switch statement to handle different message types, and just drop through to an 'Unsupported Message Type' sort of error as the default case?

dan-jones-iw: yea would be good to reduce the amount of places i have to change for a new message type :+1:

mfjarvis (approved):

PR #355, opened 2021-02-11
 AEA-1121 Add tests to update prescriptions and Add cancelation smoke-tests back 

booshibooshi:


tests/e2e/pact/specs/live/process.successes.spec.ts
```diff
@@ -34,14 +34,6 @@ const orderUpdatePactGroups = [
     name: "secondarycare-community-acute-cancel",
     cases: TestResources.processSecondaryCareCommunityAcuteOrderUpdateCases
   },
-  {
-    name: "secondarycare-community-repeatdispensing-cancel",
```
dan-jones-iw, line 38: how come these are being removed?

booshibooshi: we don't have any example cancelations for these groups, only secondarycare-community-acute-cancel, and primarycare-cancel

scripts/update_prescriptions.py
```diff
@@ -6,93 +6,149 @@
...
+
+def get_organisation_code(prepare_request_json):
+    # secondary-care
+    for entry in prepare_request_json["entry"]:
+        resource = entry["resource"]
+        if resource["resourceType"] == "HealthcareService":
+            for identifier in resource["identifier"]:
+                return identifier["value"]
+    # primary care
+    resources = map(getResource, prepare_request_json['entry'])
```
mfjarvis, line 112: Might be easier to read if you mirrored the code for HealthcareService, e.g.:
```
for entry in prepare_request_json["entry"]:
        resource = entry["resource"]
        if resource["resourceType"] == "Organization" and partOf in resource:
            for identifier in resource["identifier"]:
                return identifier["value"]
```
Up to you though

scripts/update_prescriptions.py
```diff
@@ -169,9 +230,150 @@ def update_examples(api_base_url):
...
+            if "validityPeriod" in resource["dispenseRequest"]:
+                validityStart = resource["dispenseRequest"]["validityPeriod"]["start"]
+                validityEnd = resource["dispenseRequest"]["validityPeriod"]["end"]
+    assert validityStart == date.today().isoformat()
+    assert validityEnd == (date.today() + timedelta(weeks=4)).isoformat()
+
+
+def test_get_organisation_code__gets_org_code_from_secondary_care(secondary_care_repeat_dispensing_process_request):
+    organisationCode = get_organisation_code(secondary_care_repeat_dispensing_process_request)
+    for entry in secondary_care_repeat_dispensing_process_request['entry']:
```
mfjarvis, line 353: The code for the test duplicates the code being tested. Same for the primary care equivalent. If we know which message we're going to use for the test, can we just put the expected value in the test as a literal?

mfjarvis (approved):

scripts/update_prescriptions.py
booshibooshi: I think we need to handle 2 use-cases: primary care with 2 orgs where one is main and also where only 1 org is given, so we're ordering the list by main org last to handle both in one, but if there's a way of writing that so it's clearer it'd be good to replace 👍

scripts/update_prescriptions.py
booshibooshi: We still have to traverse the python dict to read the values as we're checking that they were updated to new values so that code ends up being almost the same code as the update itself

scripts/update_prescriptions.py
mfjarvis: In both cases, the more granular of the two Organizations will have a `partOf`. The `partOf` bit can can either contain the organisation details or a reference to an entire Organization resource

scripts/update_prescriptions.py
mfjarvis: Ah yeah forgot our examples aren't static. Which is odd given the code I'm reviewing...

mfjarvis (approved):

PR #365, opened 2021-02-18
AEA-1143 - Add dispensing TBC and update spec

tjhmoore:


specification/electronic-prescription-service-api.yaml
```diff
@@ -63,33 +56,48 @@ info:
...
 
     If you have any other queries, please [contact us](https://digital.nhs.uk/developer/help-and-support).
+    
+    ### Prescribing API
+    Prescribing flows for EPS FHIR API are now in a state of technical go-live awaiting our first of type integration.
+
+    The API is currently available for sandbox testing and integration testing, but not for production use.
+
+    ### Dispensing API
+    Dispensing flows for EPS FHIR API are in the sandbox environment only.
```
dan-jones-iw, line 100: only some parts of dispensing are in sandbox - withdraw and return are spec only

specification/electronic-prescription-service-api.yaml
```diff
@@ -63,33 +56,48 @@ info:
...
+    ### To download a prescription to be dispensed: 
     1. POST a [Task-release-message](https://simplifier.net/guide/DigitalMedicines/release-duplicate-2) to the `/FHIR/R4/Task/$release` endpoint to mark the prescription as released
     2. POST a [prescription-dispense-notification](https://simplifier.net/guide/DigitalMedicines/prescription-dispense) message to the `/FHIR/R4/$process-message` endpoint to inform SPINE of a partial/successful dispense
 
-    ## Legal Use
-    This API can only be used where there is a legal basis to do so. Make sure you have a valid use case before you go too far with your development.
-
-    You must do this before you can go live (see ‘Onboarding’ below).
+    ### To notify Spine that a prescription has been dispensed:
+    1. TBC
```
dan-jones-iw, line 69: this is point 2 of the above

specification/electronic-prescription-service-api.yaml
```diff
@@ -63,33 +56,47 @@ info:
...
 
     If you have any other queries, please [contact us](https://digital.nhs.uk/developer/help-and-support).
+    
+    ### Prescribing API
+    Prescribing flows for EPS FHIR API are now in a state of technical go-live awaiting our first of type integration.
+
+    The API is currently available for sandbox testing and integration testing, but not for production use.
+
+    ### Dispensing API
+    Download and the mark as dispensed dispensing flows for the EPS FHIR API are in the sandbox environment only, withdraw and return and currently only referenced in this spec.
```
dan-jones-iw, line 99: update to say `withdraw and return are currently`

dan-jones-iw: i think Tony would like these 'spec only' interactions to still be referenced at the top of the document in the you can: section

dan-jones-iw: so would say something like 
```
you can: 
- return a prescription after downloading (spec only)
- same with withdraw

dan-jones-iw (approved):

PR #369, opened 2021-02-19
Testing tool updates

booshibooshi:


tests/e2e/pact/broker/verify.ts
```diff
@@ -93,40 +87,43 @@ async function verifyConvert(): Promise<any> {
...
-  pactGroup = ""
-  resetBackOffRetryTimer()
-  await verifyWith2Retries()
-}
-
-function isSandbox() {
-  return process.env.PACT_PROVIDER_URL.includes("sandbox")
+    await getPreparePactGroups().reduce(async (promise, group) => {
+      await promise
+      endpoint = "prepare"
```
mfjarvis, line 92: Is there a reason for `endpoint` and `pactGroup` to be top level variables instead of just being passed into `verifyX()`?

tests/e2e/pact/resources/common.ts
```diff
@@ -6,30 +6,103 @@ import {ExampleFile} from "../models/files/example-file"
...
 
+// get pact groups for verification
+export const pactGroupNames = convertPactDescriptionsToPactNames(pactGroups)
+const releasePactGroupNames = convertPactDescriptionsToPactNames(releasePactGroups)
+const cancelPactGroupNames = convertPactDescriptionsToPactNames(cancelPactGroups)
+
+// convert pact group name from description search string format to single string
+// matching the published pact's name
+function convertPactDescriptionsToPactNames(descriptions: readonly string[]) {
+ return descriptions.map(g => g.replace(/-/g, "").replace(/\s/g, "-"))
```
mfjarvis, line 74: Might be good to create a method which does this replacement and use it in both places (here and line 56)

tests/e2e/pact/resources/test-resources.ts
```diff
@@ -24,13 +25,14 @@ function getProcessCases(searchString: string, operation: string) {
...
-export const processSecondaryCareCommunityAcuteOrderCases = getProcessCases("secondary-care community acute", "send")
-export const processSecondaryCareCommunityAcuteOrderUpdateCases = getProcessCases("secondary-care community acute", "cancel")
-export const processSecondaryCareCommunityRepeatDispensingOrderCases = getProcessCases("secondary-care community repeat-dispensing", "send")
-export const processSecondaryCareCommunityRepeatDispensingOrderUpdateCases = getProcessCases("secondary-care community repeat-dispensing", "cancel")
-export const processSecondaryCareHomecareOrderCases = getProcessCases("secondary-care homecare", "send")
-export const processSecondaryCareHomecareOrderUpdateCases = getProcessCases("secondary-care homecare", "cancel")
-export const processPrimaryCareOrderCases = getProcessCases("primary-care", "send")
-export const processPrimaryCareOrderUpdateCases = getProcessCases("primary-care", "cancel")
+
+export const convertCaseGroups = pactGroups.map(pactGroup => new PactGroupCases(pactGroup, getConvertCases(pactGroup)))
```
mfjarvis, line 29: I think this might be clearer if we created the `PactGroupCases` wrapper inside the `getXCases()` functions

mfjarvis: Also why are we using `pactGroups` here instead of `getConvertPactGroups()`? Same for the other groups of groups

make.ps1
```diff
@@ -63,6 +63,7 @@ function create-smoke-tests() {
     Remove-Item Env:\LOG_LEVEL -ErrorAction SilentlyContinue
     cd tests/e2e/pact
     Remove-Item './pact' -Recurse -ErrorAction SilentlyContinue
+    npm run clear-cache
```
mfjarvis, line 66: What does this do and why do we need it?

tests/e2e/pact/package.json
```diff
@@ -6,11 +6,12 @@
...
     "lint": "eslint . --ext .ts --max-warnings 0 --fix",
     "check-licenses": "node_modules/.bin/license-checker --failOn GPL --failOn LGPL",
-    "generate-postman-collection": "ts-node postman-pact.ts --use_strict",
+    "generate-postman-collection": "ts-node postman-pact.ts --use_strict --transpile-only",
+    "clear-cache": "jest --clearCache",
     "create-sandbox-pacts": "jest sandbox",
     "create-live-pacts": "jest live",
-    "publish-pacts": "ts-node ./broker/publish.ts --use_strict",
-    "verify-pacts": "ts-node ./broker/verify.ts --use_strict"
+    "publish-pacts": "ts-node ./broker/publish.ts --use_strict --transpile-only",
```
mfjarvis, line 13: What's the reason for adding this flag?

tests/e2e/pact/postman-pact.ts
```diff
@@ -53,5 +59,7 @@ function createPostmanItem(interaction) {
         item.request.header.push({key, value})
     }
 
+    item.request.header.push({key: "Authorization", value: `Bearer ${process.env.APIGEE_ACCESS_TOKEN}`})
```
mfjarvis, line 62: Not sure how much use this is given it'll expire after 10 minutes, but I guess it shows you where to put the token?

tests/e2e/pact/broker/verify.ts
booshibooshi: Leftover from refactoring, changed to pass variables in

tests/e2e/pact/resources/common.ts
booshibooshi: Good shout

make.ps1
booshibooshi: This runs npm target clear-cache in package.json, it runs jest --clearCache, this resolves an issue where if you're making heavy refactoring changes like deleting/renaming specs jest still attempts to run them even though they no longer exist

tests/e2e/pact/package.json
booshibooshi: This was an attempt to fix the issue described above, solved by adding jest --clearCache instead, have now removed this flag

tests/e2e/pact/postman-pact.ts
booshibooshi: Not amazingly useful, but more useful than not having a token, if you run the tools to generate new prescriptions and you want to send a few into int then you can do that easily enough within the time, also gets overidden if you set postman authorization header so as a default it's ok

tests/e2e/pact/resources/test-resources.ts
booshibooshi: pactGroups vs get..PactGroups returns PactGroup type vs string respectively, the first is used for type safety and grouping by matching description, the second is used for the pact provider name so strips out spaces and formats for readability

mfjarvis (approved):

PR #370, opened 2021-02-19
Aea 990 unique medication identifiers

conner-jowitt:


coordinator/tests/services/validation/bundle-validator.spec.ts
```diff
@@ -186,6 +186,28 @@ describe("MedicationRequest consistency checks", () => {
 
     validateValidationErrors(validationErrors)
   })
+
+  test("Should reject message where 2 or more medication requests share an identifier", () => {
+    const identifier = [
```
dan-jones-iw, line 191: can define type like `const identifier: Array<fhir.Identifier> = ` rather than forcing it at the end

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -87,8 +82,16 @@ export function verifyPrescriptionBundle(bundle: fhir.Bundle): Array<errors.Vali
   const isRepeatDispensing = courseOfTherapyTypeCode === fhir.CourseOfTherapyTypeCode.CONTINUOUS_REPEAT_DISPENSING
   const repeatDispensingErrors = isRepeatDispensing ? verifyRepeatDispensingPrescription(medicationRequests) : []
 
+  const otherErrors = []
```
dan-jones-iw, line 85: feel like the name `incorrectValueErrors` was more descript

conner-jowitt: didn't think it matched with a non-unique value error, so I thought this was better suited as a holder for single error types (instead of the other holders that check a bunch of stuff)

conner-jowitt: maybe `invalidValueErrors`?

conner-jowitt: but then I guess anything that throws an error is an invalid value

dan-jones-iw: yea lol thats true but still more accurate than `other`

dan-jones-iw (approved):

PR #372, opened 2021-02-19
AEA-1037: Review changes. Tidying up.

mfjarvis:


coordinator/src/models/fhir/medication-request.ts
```diff
@@ -11,17 +10,17 @@ export enum CourseOfTherapyTypeCode {
...
     "Short course (acute) therapy"
   ),
-  CONTINUOUS: createCodeableConcept(
+  CONTINUOUS: common.createCodeableConcept(
     "http://terminology.hl7.org/CodeSystem/medicationrequest-course-of-therapy",
     CourseOfTherapyTypeCode.CONTINUOUS,
     "Continuous long term therapy"
   ),
-  CONTINOUS_REPEAT_DISPENSING: createCodeableConcept(
+  CONTINOUS_REPEAT_DISPENSING: common.createCodeableConcept(
```
dan-jones-iw, line 23: `CONTINUOUS_`

coordinator/src/models/fhir/medication-request.ts
```diff
@@ -11,17 +10,17 @@ export enum CourseOfTherapyTypeCode {
 }
 
 export const CourseOfTherapyType = Object.freeze({
```
dan-jones-iw, line 12: seems like `Object.freeze` doesnt stop me from changing values within the codeable concept. for this to be useful wouldnt we like it to stop any changes at all?

mfjarvis: It doesn't but neither would
```
const SOME_CONSTANT = {
  key1: "value1",
  key2: "value2"
}
```
so it's as "immutable" as any object constant

mfjarvis: Replaced with equivalent top level consts

coordinator/src/models/fhir/medication-request.ts
mfjarvis: Fixed

dan-jones-iw (approved):

PR #389, opened 2021-03-10
Aea 1201 Handle Unknown CPA IDs

conner-jowitt:


coordinator/src/services/communication/live-spine-client.ts
```diff
@@ -35,8 +36,20 @@ export class LiveSpineClient implements SpineClient {
...
 
   async send(spineRequest: SpineRequest, logger: Logger): Promise<SpineResponse<unknown>> {
-    logger.info("Building EBXML wrapper for SpineRequest")
-    const wrappedMessage = this.ebXMLBuilder(spineRequest, logger)
+    let wrappedMessage
+    try {
+      logger.info("Building EBXML wrapper for SpineRequest")
+      wrappedMessage = this.ebXMLBuilder(spineRequest, logger)
+    } catch (error) {
+      if (error instanceof FhirMessageProcessingError) {
```
mfjarvis, line 44: I'm not sure you need to turn this into an OperationOutcome here as we have a thing in app.ts which does that already. Could just let the exception get thrown all the way up

mfjarvis (approved):

PR #391, opened 2021-03-12
[WIP] AEA-1128 add sandbox responses for task withdraw and return

conner-jowitt:


coordinator/src/models/errors/validation-errors.ts
```diff
@@ -84,3 +84,27 @@ export function createResourceTypeIssue(expectedResourceType: string): fhir.Oper
...
 }
+
+export function createTaskIncorrectValueIssue(
+  fieldName: string,
+  requiredFieldValue: string
+): fhir.OperationOutcomeIssue {
+  return {
+    severity: "error",
+    code: "value",
+    diagnostics: `Task.${fieldName} must be '${requiredFieldValue}'.`,
```
mfjarvis, line 95: This message won't look right when requiredFieldValue is "in-progress or rejected" because of the quotes. Could just remove the quotes, or could accept an array of acceptable field values

coordinator/src/services/validation/task-validator.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  return validationErrors
+}
+
+function checkValidSystem(task: fhir.Task, system: string): Array<fhir.OperationOutcomeIssue>{
+  const validSystemCode = getCodingForSystemOrNull(
+    task.reasonCode.coding,
+    system,
+    `task.reasonCode`
```
mfjarvis, line 45: FHIR path should be `Task.reasonCode` with an uppercase T

coordinator/src/models/errors/validation-errors.ts
conner-jowitt: I guess remove the quotes and pass quotes in to the function? `must be 'in-progress' or 'rejected'` is probably the best way to display it

coordinator/src/services/validation/task-validator.ts
conner-jowitt: Will change :+1:

mfjarvis: Can use double quotes as well now

coordinator/src/services/validation/task-validator.ts
```diff
@@ -0,0 +1,54 @@
...
+  if(task.intent != "order") {
+    validationErrors.push(errors.createTaskIncorrectValueIssue("intent", "order"))
+  }
+
+  if(task.status === "in-progress") {
+    if(!task.code){
+      validationErrors.push({
+        severity: "error",
+        code: "value",
+        diagnostics: `Task.code is requred when task.status="in-progress" .`,
```
mfjarvis, line 21: Extra space at the end + can use double quotes

mfjarvis (dismissed):

mfjarvis (approved):

PR #394, opened 2021-03-12
AEA-1131 Enrich dispense notification from ods

booshibooshi:


coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -170,6 +172,10 @@ export function verifyDispenseBundle(bundle: fhir.Bundle): Array<fhir.OperationO
     }
   })
 
+  if (medicationDispenses.some(medicationDispense => !getOrganisationPerformer(medicationDispense))) {
+    allErrors.push(errors.createMedicationDispenseMissingValueIssue("performer.actor"))
```
booshibooshi, line 176: fhir path performer.actor is not clear feedback to user that this is because there is a performer actor type of organization missing here. 

To confirm:
1) Mapping is correct
2) validation for this is necessary outside of fhir validator (if not, remove this check, if yes then make error clearer)
3) IG states we need Organization

booshibooshi: Have made this error feedback clearer for the moment, to discuss the solution for this in another ticket

coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
```diff
@@ -271,26 +250,7 @@ describe("fhir MedicationDispense maps correct values in DispenseNotificiation",
     })
   })
 
-  // eslint-disable-next-line max-len
-  test("performer.actor.(type === 'Practitioner') maps to pertinentInformation1.pertinentSupplyHeader.author.AgentPerson.agentPerson", () => {
-    medicationDispenses.forEach(medicationDispense => setPractitionerName(medicationDispense, "XX-TEST-VALUE"))
```
booshibooshi, line 276: This has been migrated to using unattended access user which is using ods lookup, we have tests surrounding lookups for ods already

coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
```diff
@@ -109,59 +111,10 @@ describe("fhir MessageHeader maps correct values in DispenseNotificiation", () =
     messageHeader = getMessageHeader(dispenseNotification)
   })
 
-  test("destination.receiver.identifier maps to primaryInformationRecipient.AgentOrg.agentOrganization", () => {
-    const fhirMessageDestination = onlyElement(messageHeader.destination, "MessageHeader.destination")
```
booshibooshi, line 113: mapping has moved to MedicationDispense.performer.actor

coordinator/src/services/translation/request/return/return.ts
```diff
@@ -14,7 +13,8 @@ export async function convertTaskToDispenseProposalReturn(
   task: fhir.Task,
   logger: pino.Logger
 ): Promise<hl7V3.DispenseProposalReturn> {
-  const id = createIdFromTaskIdentifier(task.identifier)
+  const idValue = getMessageId(task.identifier)
```
mfjarvis, line 16: I think we'll need to pass in the FHIR path if we do this. Otherwise any error messages will refer to `Bundle.identifier`

coordinator/src/services/translation/request/send-message-payload.ts
```diff
@@ -73,14 +73,9 @@ function convertRequesterToControlActAuthor(
   bundle: fhir.Bundle
 ) {
 
-  // todo dispenseNotification: implement dispense verson
   const messageType = identifyMessageType(bundle)
   if (messageType === fhir.EventCodingCode.DISPENSE) {
```
mfjarvis, line 77: If dispense notifications should use the unattended author details, should the translations just call `createControlActEventForUnattendedAccess()` instead of `createControlActEvent()`? Then we can remove this bespoke logic.

mfjarvis (approved):

PR #395, opened 2021-03-14
make use of auth level header in prescription journey

dan-jones-iw:


tests/e2e/pact/services/get-access-token.ts
```diff
@@ -0,0 +1,36 @@
+import {createJWT} from "./jwt"
+import axios from "axios"
+import {URLSearchParams} from "url"
+import * as fs from "fs"
+import path from "path"
+
+/* TODO - get from ADO / env variables */
+const internalDevApiKey = "4PD6IBEDrC4oVlQOJ5VSqk951SAqTA1G"
```
booshibooshi, line 8: is this todo required for this pr?

booshibooshi: +assuming from the location of this file we're planning on adding a smoke-test in future?

coordinator/src/routes/debug/convert.ts
```diff
@@ -25,7 +25,7 @@ export default [
       const payload = getPayload(request) as fhir.Resource
       const requestId = request.headers["nhsd-request-id"].toUpperCase()
       if (isBundle(payload)) {
-        const issues = bundleValidator.verifyBundle(payload)
+        const issues = bundleValidator.verifyBundle(payload, request.headers)
```
booshibooshi, line 28: Maybe have a verifyAuth helper method separately instead? Attaching to bundle validator feels a bit forced, i.e. auth isn't an issue with the bundle itself

booshibooshi: Also reading the implementation for the auth verification I think we'll need this for task endpoint too, as it is checking whether user/app auth is valid so we'll want those checks for task too if we want to support unattended and attended in future?

tests/e2e/pact/services/get-access-token.ts
dan-jones-iw: yea smoke tests are for a future ticket - felt like this was getting increasingly bloated so cut it off before any smoke test related things happened

dan-jones-iw: you can get an access token locally for now and test using that manually? (only in internal-dev)

coordinator/src/routes/debug/convert.ts
dan-jones-iw: thats true however the bundle validator is where we check whether the incoming bundle is part of the prescription flow or the dispense flow (and so requires different auth). in terms of task endpoints, everything atm is assumed app restricted unless it has the user auth verification on it. maybe could use some refactoring to make more clear?

coordinator/src/services/validation/auth-level.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+  return true
+}
+
+function requiresAuth() {
+  return process.env.SANDBOX !== "1"
+}
+
+function requestHasAppAuth(headers: Hapi.Util.Dictionary<string>): boolean {
+  return headers["nhsd-identity-authentication-method"]?.includes("application")
```
mfjarvis, line 19: In the Signing Service we've used this same header name to carry the AMR value, which is a similar but different thing. I think we should call this something else to avoid confusion

coordinator/src/routes/debug/convert.ts
booshibooshi: Yeah I think some refactoring here would help clear things up

coordinator/src/services/validation/auth-level.ts
dan-jones-iw: ive gone with `NHSD-Identity-Authentication-Level`

booshibooshi (dismissed):

coordinator/src/routes/process.ts
```diff
@@ -13,6 +23,13 @@ export default [
     method: "POST",
     path: `${BASE_PATH}/$process-message`,
     handler: async (request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit): Promise<Hapi.ResponseObject> => {
+      const hopefullyBundle = getPayload(request) as fhir.Bundle
```
mfjarvis, line 26: We already do this on line 38, just move it higher up

mfjarvis (dismissed):

booshibooshi (approved):

PR #396, opened 2021-03-15
Unit tests for Task validator

conner-jowitt:


dan-jones-iw (dismissed):

coordinator/tests/resources/test-resources.ts
```diff
@@ -105,6 +105,18 @@ export const specification = [
   examplePrescription3
 ]
 
+export const exampleWithdrawTask: fhir.Task= JSON.parse(fs.readFileSync(
+  // eslint-disable-next-line max-len
+  path.join(__dirname, "./secondary-care/community/acute/nominated-pharmacy/clinical-practitioner/1-Task-Request-Withdraw-200_OK.json"),
```
ri-ch, line 110: Id move the root of this filepath out to a const to we dont need to suppress the eslint warning

conner-jowitt: The lines are still too long, even with it pulled out. I could pull it out into 2 strings and stick them together?

dan-jones-iw (approved):

dan-jones-iw (approved):

dan-jones-iw (approved):

PR #398, opened 2021-03-16
AEA-1125: Tidy up shared translation code.

mfjarvis:


coordinator/src/services/translation/request/index.ts
```diff
@@ -60,10 +70,11 @@ export async function createDispenseNotificationSendMessagePayload(
 ): Promise<hl7V3.SendMessagePayload<hl7V3.DispenseNotificationRoot>> {
   const dispenseNotification = await convertDispenseNotification(bundle, logger)
   const dispenseNotificationRoot = new hl7V3.DispenseNotificationRoot(dispenseNotification)
-  return createSendMessagePayloadForUnattendedAccess(
+  return createSendMessagePayload(
+    getMessageIdFromBundle(bundle),
     hl7V3.Hl7InteractionIdentifier.DISPENSE_NOTIFICATION,
-    dispenseNotificationRoot,
-    getMessageId([bundle.identifier], "Bundle.identifier")
+    undefined,
```
booshibooshi, line 76: Can we move the responsibility to construct the author into a intermediate function so we don't have to know whether this should be undefined or not when calling this method. Something like...

`export function createCancellationSendMessagePayload(
  bundle: fhir.Bundle
): hl7V3.SendMessagePayload<hl7V3.CancellationRequestRoot> {
  const cancellationRequest = convertCancellation(bundle)
  const cancellationRequestRoot = new hl7V3.CancellationRequestRoot(cancellationRequest)
  const interactionId = hl7V3.Hl7InteractionIdentifier.CANCEL_REQUEST
  return createSendMessagePayload(
    getMessageIdFromBundle(bundle),
    getAuthor(interactionId, bundle),
    convertResponsiblePractitionerToControlActAuthor(bundle),
    cancellationRequestRoot
  )
}

function getAuthor(interaction: hl7V3.Hl7InteractionIdentifier, request: fhir.Bundle | fhir.Task): any {
  if (interaction === hl7V3.Hl7InteractionIdentifier.PARENT_PRESCRIPTION_URGENT
    || hl7V3.Hl7InteractionIdentifier.CANCEL_REQUEST) 
  {
    return convertResponsiblePractitionerToControlActAuthor(request as fhir.Bundle)
  }
  return undefined
}`

mfjarvis: I think that might make it harder to see what's going on. Happy to create a `createSendMessagePayloadUnattended()` which passes in `undefined` for the author though. I'm also keeping in mind that some of these unattended interactions might also support attended use, in which case we could have an author depending on the access mode used

booshibooshi: Sounds good, just want to avoid having to know to pass in undefined as a parameter without context, having the methods split between createSendMessagePayload and createSendMessagePayloadUnattended sounds good as it's providing meaning as to why they're different and solves the undefined problem

booshibooshi: Wrapper method for selecting unattended or attended createSenessage... method would be fine in the scenario you mentioned above. But passing undefined is much easier to accidentally make a big I think

booshibooshi (approved):

PR #411, opened 2021-03-26
AEA-1160 Fix update_prescriptions + allow skipping broker for smoke-tests

booshibooshi:


make.ps1
```diff
@@ -62,15 +59,16 @@ function create-smoke-tests() {
     . ./envrc.ps1
     $env:PACT_VERSION="$env:USERNAME".replace(' ','')
     $env:UPDATE_PRESCRIPTIONS=$update
-    $env:SIGNING_PRIVATE_KEY_PATH="../../../sign/eps_int_test_private.key"
-    $env:SIGNING_CERT_PATH="../../../sign/eps_int_test_certificate.crt"
+    $env:PACT_USE_BROKER=$broker
```
mfjarvis, line 62: Does the equivalent need adding to the makefile too?

ri-ch, line 62: Were we not going to remove the broker entirely for now? (We could probably do this as a new PR)

ri-ch (approved):

make.ps1
booshibooshi: I think we're dependant on it at the moment for both debugging failing smoke tests since the IDs and signatures are generated on each run and not persisted anywhere else and it's currently fulfilling the function of test reports. So I think we'll need it for a little longer till we have other solutions in place to address both of those points

booshibooshi: Will raise a new pr

PR #414, opened 2021-03-26
AEA-1160 Fix error in bundle validator dispense-notification test

booshibooshi:


coordinator/tests/services/validation/bundle-validator.spec.ts
```diff
@@ -352,7 +352,7 @@ describe("verifyDispenseNotificationBundle", () => {
 
   test("returns an error when MedicationDispenses have different performer values per type", () => {
     const medicationDispenseEntry =
-      bundle.entry.filter(entry => entry.resource.resourceType === "MedicationDispense")[0]
+      clone(bundle.entry.filter(entry => entry.resource.resourceType === "MedicationDispense")[0])
```
dan-jones-iw, line 355: can we use the `getMedicationDispenses` here?

dan-jones-iw (approved):

coordinator/tests/services/validation/bundle-validator.spec.ts
booshibooshi: Not quite, it's one level above, getting the whole entry rather than the resource itself 👍

PR #418, opened 2021-03-30
Move models from coordinator to models module

booshibooshi:


coordinator/tests/resources/test-resources.ts
```diff
@@ -3,9 +3,8 @@ import {ElementCompact} from "xml-js"
 import * as fs from "fs"
 import * as path from "path"
 import * as LosslessJson from "lossless-json"
-import {SpineDirectResponse} from "../../src/models/spine"
-import * as hl7V3 from "../../src/models/hl7-v3"
-import {fhir, fetcher} from "@models"
+import {} from "@models"
```
mfjarvis, line 6: This one can go

models/hl7-v3/core.ts
```diff
@@ -172,3 +171,16 @@ export class IntervalUnanchored {
     }
   }
 }
+
+function getNumericValueAsString(numericValue: string | number | LosslessNumber): string {
```
mfjarvis, line 175: This is for use with parsed FHIR messages, so probably belongs elsewhere

models/hl7-v3/dispense-notification.ts
```diff
@@ -38,7 +36,7 @@ export class DispenseNotification implements ElementCompact {
       "163541000000107",
       "Dispensed Medication - FocusActOrEvent (administrative concept)"
     )
-    this.effectiveTime = convertMomentToHl7V3DateTime(moment.utc())
+    this.effectiveTime = new core.Timestamp("PLACEHOLDER")
```
mfjarvis, line 39: What's the reason for putting this back?

models/hl7-v3/core.ts
booshibooshi: yeah agreed, doesn't really fit into `models` but it was being used by models, I'll remove this from the NumericValue constructor, set it to take in a string and set a string, and see what the impact is

models/hl7-v3/dispense-notification.ts
booshibooshi: because this is now in models and the `convertMomentToHl7V3DateTime` function is in `coordinator`, so it's constructed as a placeholder to avoid circular dependency and then actually set in the translation in coordinator

models/hl7-v3/core.ts
booshibooshi: removed from constructor, no impact

mfjarvis (approved):

PR #419, opened 2021-03-30
aea-964

dan-jones-iw:
update mapping of effectiveTime to be translated to prescriptionStatus extension in cancel response

### DIFF
--- a/coordinator/src/services/translation/response/cancellation/cancellation-medication-request.ts
+++ b/coordinator/src/services/translation/response/cancellation/cancellation-medication-request.ts
@@ -39,7 +39,7 @@ export function createMedicationRequest(
     intent: fhir.MedicationRequestIntent.ORDER,
     medicationCodeableConcept: MEDICINAL_PRODUCT_CODEABLE_CONCEPT,
     subject: fhir.createReference(patientId),
-    authoredOn: undefined,
+    authoredOn: undefined, //the v3 message doesnt have enough information for authoredOn
     requester: fhir.createReference(originalPrescriptionAuthorPractitionerRoleId),
     groupIdentifier: createGroupIdentifierFromPertinentInformation2(cancellationResponse.pertinentInformation2)
   }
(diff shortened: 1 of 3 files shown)


coordinator/src/services/translation/response/cancellation/cancellation-medication-request.ts
```diff
@@ -31,32 +32,36 @@ export function createMedicationRequest(
...
+      effectiveTime
     ),
     identifier: createItemNumberIdentifier(cancellationResponse.pertinentInformation1),
     status: medicationRequestStatus,
     intent: fhir.MedicationRequestIntent.ORDER,
     medicationCodeableConcept: MEDICINAL_PRODUCT_CODEABLE_CONCEPT,
     subject: fhir.createReference(patientId),
-    //TODO - effectiveTime should probably be the timestamp of the status, not authoredOn
-    authoredOn: convertHL7V3DateTimeToIsoDateTimeString(cancellationResponse.effectiveTime),
+    authoredOn: undefined,
```
ri-ch, line 42: Do we really want this to be undefined? Should it even be in the model if we're not going to apply a value to it?

dan-jones-iw: i left it as undefined since i wanted it to be explicitly clear that the v3 message doesnt have enough information for authoredOn. this will never change so happy to delete the line. its in the model since the incoming FHIR medication request has authoredOn as a field

booshibooshi (dismissed):

coordinator/src/services/translation/response/cancellation/cancellation-medication-request.ts
ri-ch: OK, that makes sense. Just put a summarised version of that in the code as a comment. Presumably the serialiser will just remove the field altogether?

ri-ch (approved):

PR #424, opened 2021-03-31
Remove tests from bundled coordinator module

booshibooshi:


coordinator/tsconfig.json
```diff
@@ -15,18 +15,16 @@
...
-        "tests/*"
+        "node_modules/*"
       ],
-      "@models": ["../models", "./models"]
+      "@models": ["../models"]
     }
   },
   "include": [
-    "src/**/*",
-    "tests/**/*"
```
mfjarvis, line 26: We added tests to this file for a reason - this allows the IDE to know that our tsconfig applies to our tests, and so it doesn't complain about things like `flatMap` not existing on arrays. Only the IDE uses this file. The build uses the other tsconfig file which excludes tests

coordinator/tsconfig-build.json
```diff
@@ -1,6 +1,3 @@
 {
-  "extends": "./tsconfig.json",
-  "include": [
-    "src/**/*"
-  ]
+  "extends": "./tsconfig.json"
```
mfjarvis, line 2: Doing this makes this file pointless. See below

mfjarvis: (the thing you removed was overriding the other file to exclude tests)

mfjarvis:
Why does the e2e tests module use coordinator in the first place? Shouldn't shared code be in the common module?

mfjarvis:
> Why does the e2e tests module use coordinator in the first place? Shouldn't shared code be in the common module?

I've answered my own question - it's because we sign the prescriptions in the e2e tests.

PR #427, opened 2021-04-07
Cleanup smoke tests

booshibooshi:


tests/e2e/pact/specs/sandbox/prepare.successes.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+            path: apiPath,
+            body: JSON.parse(requestStr)
+          },
+          willRespondWith: {
+            headers: {
+              "Content-Type": "application/json",
+              "X-Request-ID": requestId,
+              "X-Correlation-ID": correlationId
+            },
+            body: {
```
dan-jones-iw, line 45: thoughts about loading the sandbox return files in the body check?

booshibooshi: We cant use exact response files as the digest and timestamp are different every time you call the api due to the timestamp being added per api call and this forming part of the digest as well as returned in a field.

We are using the sandbox responses to generate examples of what these fields look like and the below code ensures that the body will be identical except for digest and timestamp values which are checked that they contain any string value

dan-jones-iw: sorry yea just realised ive been an absolute clown, the sandbox responses are in v3 anyway

dan-jones-iw (dismissed):

dan-jones-iw (approved):

PR #428, opened 2021-04-09
Translate Error Codes

conner-jowitt:


coordinator/tests/resources/test-resources.ts
```diff
@@ -169,6 +173,7 @@ const syncMultipleError: ExampleSpineResponse = {
     statusCode: 400
   },
   spineErrorCode: "202",
+  epsErrorCode: "ERROR",
```
booshibooshi, line 176: 👍

booshibooshi (commented):
ACs state we were adding a couple of smoke tests, are these still todo?

dan-jones-iw (approved):

PR #430, opened 2021-04-13
validate endpoint

dan-jones-iw:


tests/e2e/pact/specs/live/validate.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+          uponReceiving: "a valid FHIR message",
+          withRequest: {
+            headers: {
+              "Content-Type": "application/fhir+json; fhirVersion=4.0",
+              "X-Request-ID": requestId,
+              "X-Correlation-ID": correlationId
+            },
+            method: "POST",
+            path: apiPath,
+            body: LosslessJson.parse(messageStr)
```
booshibooshi, line 38: Will need to be JSON parse here, as lossless parse adds extra fields resulting in mismatch between expected and actual

booshibooshi (commented):
Verify.ts will also need updating to verify the validate interaction, at the moment the pact will be created but not verified

tests/e2e/pact/specs/live/validate.spec.ts
dan-jones-iw: just saw the failure for this lol, fixing

booshibooshi (approved):

PR #432, opened 2021-04-14
Handle missing/optional extensions in bundle validation

booshibooshi:


coordinator/tests/services/translation/request/line-item.spec.ts
```diff
@@ -22,13 +22,27 @@ describe("convertMedicationRequestToLineItem", () => {
...
   })
 
+  test("Throws TooFewValuesError when passed no dispenseRequest performer-site-type extension", () => {
+    firstFhirMedicationRequest.dispenseRequest.extension = []
+    expect(() =>
+      convertBundleToPrescription(bundle)
+    ).toThrow(errors.TooFewValuesError)
+  })
+
+  test("Throws InvalidValueError when passed no dispenseRequest extensions", () => {
```
dan-jones-iw, line 39: tried sending an example of this in and got a generic 500 back, is this error being hidden by the proxy?

dan-jones-iw (approved):

coordinator/tests/services/translation/request/line-item.spec.ts
booshibooshi: Confirmed that we are receiving 400 for this scenario

PR #434, opened 2021-04-16
Test production releases

booshibooshi:


manifest_template.yml
```diff
@@ -20,6 +20,12 @@ APIGEE_ENVIRONMENTS:
   name: int
   ratelimit: 20ps
   quota: 1200
+- display_name: Production
+  name: prod
+  approval_type: manual
+  ratelimit: '15000pm'        # todo: confirm with Spine/EPS team what these should be
+  quota: '7800'               # these values match PDS
```
dan-jones-iw, line 27: should these be in '? just comparing to above

manifest_template.yml
```diff
@@ -20,6 +20,12 @@ APIGEE_ENVIRONMENTS:
   name: int
   ratelimit: 20ps
   quota: 1200
+- display_name: Production
+  name: prod
+  approval_type: manual
+  ratelimit: '15000pm'        # todo: confirm with Spine/EPS team what these should be
```
mfjarvis, line 26: For now I'd probably match what we're using for int. 20 tps should be more than enough to start with

dan-jones-iw (approved):

PR #438, opened 2021-04-20
Update electronic-prescription-service-api.yaml

SachinPath:
Updated Overview section of spec for better clarity.


specification/electronic-prescription-service-api.yaml
```diff
@@ -371,7 +372,7 @@ paths:
   /FHIR/R4/Task/$release:
     post:
       operationId: release
-      summary: Send a prescription release request to EPS
+      summary: Send a download prescription request to EPS
```
SachinPath, line 375: Should be same as you can section.

PR #449, opened 2021-05-05
AEA-1356: Added dispense error mappings, updated tests to check defau…

callumtipper-IW:
…lt and specific error cases


coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -288,6 +377,17 @@ export class CancelResponseHandler extends SpineResponseHandler<hl7V3.Cancellati
   }
 }
 
+export class ReleaseRejectionHandler extends SpineResponseHandler<hl7V3.PrescriptionReleaseRejectRoot> {
+  protected extractErrorCodes(
+    sendMessagePayload: hl7V3.SendMessagePayload<hl7V3.PrescriptionReleaseRejectRoot>
+  ): Array<hl7V3.Code<string>>{
+    return [
```
mfjarvis, line 384: I'd change this to:
```
const errorCode = sendMessagePayload.ControlActEvent.subject.PrescriptionReleaseReject
        .pertinentInformation?.pertinentRejectionReason?.value
return errorCode ? [errorCode] : []
```
That way we handle a message without an error code (supposedly possible according to the MiM)

mfjarvis (approved):

PR #452, opened 2021-05-10
stop certain headers from being available in production

dan-jones-iw:


coordinator/src/services/environment.ts
```diff
@@ -0,0 +1,4 @@
+const environment = process.env.ENVIRONMENT
+
+export const isProd = (): boolean => environment === "prod"
```
ri-ch, line 3: Is there a reason these need to be functions instead of variables

coordinator/src/services/headers.ts
```diff
@@ -0,0 +1,27 @@
...
+export const invalidProdHeaders: Array<RequestHeaders> = [RequestHeaders.RAW_RESPONSE, RequestHeaders.SKIP_VALIDATION]
+
+export const rejectInvalidProdHeaders: Hapi.Lifecycle.Method = (
+  request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit
+) => {
+  if (isProd()) {
+    const presentInvalidHeaders = invalidProdHeaders.filter(invalidHeader => request.headers[invalidHeader])
+    if (presentInvalidHeaders) {
+      return responseToolkit
+        .response("Header invalid in production")
```
ri-ch, line 21: Not sure id refer to the environment name here. Maybe something like `Header <headername> is not supported`

coordinator/src/services/headers.ts
```diff
@@ -0,0 +1,27 @@
...
+  SKIP_VALIDATION = "x-skip-validation",
+  SMOKE_TEST = "x-smoke-test"
+}
+
+export const invalidProdHeaders: Array<RequestHeaders> = [RequestHeaders.RAW_RESPONSE, RequestHeaders.SKIP_VALIDATION]
+
+export const rejectInvalidProdHeaders: Hapi.Lifecycle.Method = (
+  request: Hapi.Request, responseToolkit: Hapi.ResponseToolkit
+) => {
+  if (isProd()) {
```
ri-ch, line 17: Can we add a log message to explain why this request was rejected?

coordinator/src/services/headers.ts
```diff
@@ -0,0 +1,27 @@
...
+    const presentInvalidHeaders = invalidProdHeaders.filter(invalidHeader => request.headers[invalidHeader])
```
ri-ch, line 18: `requestContainsInvalidHeaders`?

coordinator/tests/routes/util.spec.ts
```diff
@@ -147,7 +147,7 @@ describe("handleResponse", () => {
     server.route([createRoute(spineResponse)])
 
     const response = await server.inject(
-      createRouteOptions(spineResponse, {"X-Untranslated-Response": "true"})
+      createRouteOptions(spineResponse, {"X-Raw-Response": "true"})
```
ri-ch, line 150: Is the logic to support this in place yet?

coordinator/src/services/headers.ts
mfjarvis, line 21: I think this should be an OperationOutcome

coordinator/src/services/environment.ts
dan-jones-iw: easy way to fake prod mode in tests - mock out return value of these functions. easier than changing env variables per test

coordinator/src/services/headers.ts
dan-jones-iw: ill update the name :+1:

coordinator/tests/routes/util.spec.ts
dan-jones-iw: yep

coordinator/src/services/headers.ts
```diff
@@ -0,0 +1,36 @@
...
+) => {
+  if (isProd()) {
+    const listOfInvalidHeaders = Object.keys(request.headers).filter(
+      requestHeader => invalidProdHeaders.includes(requestHeader as RequestHeaders)
+    )
+    if (listOfInvalidHeaders.length) {
+      console.error(`Request with id: ${
+        request.headers[RequestHeaders.REQUEST_ID]
+      } had invalid header(s): ${
+        listOfInvalidHeaders.length
```
mfjarvis, line 26: Would be nicer to stringify the invalid headers rather than just including the length, but not a big deal

models/errors/validation-errors.ts
```diff
@@ -135,8 +142,21 @@ export const unauthorisedActionIssue: fhir.OperationOutcomeIssue = {
...
 }
+
+export const invalidHeaderOperationOutcome = (headers: Array<string>): fhir.OperationOutcomeIssue => ({
+  severity: "error",
+  code: fhir.IssueCodes.INVALID,
+  details: {
+    coding: [{
+      system: "https://fhir.nhs.uk/R4/CodeSystem/Spine-ErrorOrWarningCode",
+      code: "MISSING_OR_INVALID_HEADER",
+      display: "There is a required header missing or invalid"
```
mfjarvis, line 158: I'd remove the word `required` from here, given that we're also using this error when the header is forbidden

dan-jones-iw: copied this from https://fhir.nhs.uk/STU3/ValueSet/Spine-ErrorOrWarningCode-1 but might not have been the right thing to do lol

coordinator/src/services/headers.ts
ri-ch: For future ref, you dont need to include the request ID, the logger will store that for you

ri-ch (approved):

PR #455, opened 2021-05-12
add x-smoke-test to invalid prod header list and update test descriptions

dan-jones-iw:


coordinator/src/services/headers.ts
```diff
@@ -11,7 +11,9 @@ export enum RequestHeaders {
   SMOKE_TEST = "x-smoke-test"
 }
 
-export const invalidProdHeaders: Array<RequestHeaders> = [RequestHeaders.RAW_RESPONSE, RequestHeaders.SKIP_VALIDATION]
+export const invalidProdHeaders: Array<RequestHeaders> = [
+  RequestHeaders.RAW_RESPONSE, RequestHeaders.SKIP_VALIDATION, RequestHeaders.SMOKE_TEST
```
mfjarvis, line 15: Won't this stop smoke tests from working in prod?

dan-jones-iw: was listed as an invalid header on https://nhsd-jira.digital.nhs.uk/browse/AEA-1368. ill double check with rich whether this is actually what he wanted

mfjarvis (approved):

PR #458, opened 2021-05-13
AEA-1378: Fix translation of multiple prescriber endorsements.

mfjarvis:


dan-jones-iw (approved):

coordinator/src/services/translation/request/line-item.ts
```diff
@@ -36,19 +37,22 @@ export function convertPrescriptionEndorsements(
   medicationRequest: fhir.MedicationRequest,
   lineItem: hl7V3.LineItem
 ): void {
-  const endorsementExtension = getExtensionForUrlOrNull(
-    medicationRequest.extension,
-    "https://fhir.nhs.uk/StructureDefinition/Extension-DM-PrescriptionEndorsement",
-    "MedicationRequest.extension"
-  ) as fhir.CodeableConceptExtension
+  const endorsementExtensions = medicationRequest.extension?.filter(
```
ri-ch, line 40: Presumably this behaves the same way if we have 1 or _n_ endorsements?

mfjarvis: Yeah, we have smoke tests for both of those cases :+1:

PR #461, opened 2021-05-17
AEA-1383: Only publish smoke test output if smoke tests have been run.

mfjarvis:


azure/azure-release-template.yml
```diff
@@ -178,3 +178,4 @@ extends:
           pathToPublish: '$(SERVICE_NAME)/$(SERVICE_ARTIFACT_NAME)/pact/prescriptions-$(APIGEE_ENVIRONMENT).txt'
           artifactName: 'smoke-tests'
           publishLocation: 'Container'
+        condition: eq(variables['run_smoke_tests'], 'true')
```
mfjarvis, line 181: Could make this conditional on `succeeded()` as well but I thought it might be useful to see the output even if the smoke tests fail

dan-jones-iw: sounds good yea

dan-jones-iw (approved):

PR #464, opened 2021-05-17
added tests for switchContentTypeForSmokeTest server extension

dan-jones-iw:
and also refactor content types into one object for cleaner imports


callumtipper-IW (dismissed):

coordinator/src/routes/util.ts
```diff
@@ -11,10 +11,12 @@ import {RequestHeaders} from "../services/headers"
 
 type HapiPayload = string | object | Buffer | stream //eslint-disable-line @typescript-eslint/ban-types
 
-export const CONTENT_TYPE_XML = "application/xml"
-export const CONTENT_TYPE_PLAIN_TEXT = "text/plain"
-export const CONTENT_TYPE_FHIR = "application/fhir+json; fhirVersion=4.0"
-export const CONTENT_TYPE_JSON = "application/json"
+export enum contentTypes {
```
mfjarvis, line 14: Our convention has been to name enum members in upper case, and I think the enum itself would be in title case. Like this: https://google.github.io/styleguide/jsguide.html#features-objects-enums

dan-jones-iw: updated

mfjarvis (approved):

PR #465, opened 2021-05-18
Revert package/package lock changes from when instability was introduced

booshibooshi:


tests/e2e/pact/package.json
```diff
@@ -33,12 +30,15 @@
     "@types/supertest": "2.0.9",
     "@typescript-eslint/eslint-plugin": "3.2.0",
     "@typescript-eslint/parser": "3.2.0",
+    "@hapi/boom": "^9.1.0",
```
dan-jones-iw, line 33: why are these moving down :sob:

dan-jones-iw: @h is before @t in alphabet

dan-jones-iw (approved):

tests/e2e/pact/package.json
mfjarvis: Love that you just tagged two random developers in this PR :joy:

dan-jones-iw: lol oops

PR #472, opened 2021-05-21
remove userAuth handlers from /-message and /

dan-jones-iw:
endpoint names with $ are dumb


coordinator/src/routes/prescribe/prepare.ts
```diff
@@ -11,7 +14,7 @@ export default [
   {
     method: "POST",
     path: `${BASE_PATH}/$prepare`,
-    handler: userAuthValidator(externalValidator(
+    handler: externalValidator(
```
mfjarvis, line 17: Is `userAuthValidator` used anywhere now or can it just go?

dan-jones-iw: ill delete :+1:

mfjarvis (approved):

PR #473, opened 2021-05-24
AEA-1339: Populate ASID from Apigee app custom attribute.

mfjarvis:


coordinator/src/routes/prescribe/polling.ts
```diff
@@ -2,12 +2,17 @@ import {Request} from "@hapi/hapi"
...
+import {getAsidHeader} from "../../services/headers"
 
 export default [{
   method: "GET",
   path: "/_poll/{poll_path}",
   handler: async (request: Request, responseToolkit: Hapi.ResponseToolkit): Promise<Hapi.ResponseObject> => {
-    const spineResponse = await spineClient.poll(request.params.poll_path, request.logger)
+    const spineResponse = await spineClient.poll(
+      request.params.poll_path,
+      getAsidHeader(request.headers),
```
ri-ch, line 13: This could just be named `getAsid`

ri-ch (dismissed):

mfjarvis:
Holding off on merging this until we have default ASIDs in place to fall back on. Otherwise we'll break a lot of people's integrations when we release to int.

mfjarvis:
Once merged we should hold off releasing to int until RFC 30027100 has been applied.

mfjarvis:
/azp please please please run the deploy pipeline

proxies/live/apiproxy/targets/apim.xml
```diff
@@ -42,6 +42,27 @@
...
+        </Step>
+        <Step>
+          <!-- Populate the ASID var with the default ASID -->
+          <Name>KeyValueMapOperations.GetDefaultAsid</Name>
+          <Condition>(app.asid is null)</Condition>
+        </Step>
+        <Step>
+          <!-- Add the ASID var as a header -->
+          <Name>AssignMessage.AddAsidHeader</Name>
+        </Step>
```
booshibooshi, line 65: Struggled to follow this initially, so the logic is: fail if there is no FROM_ASID in the apigee app in production, for other environments try to get a FROM_ASID from apigee app but fallback to a default value in an apigee keyvaluemap which is the generic platform FROM_ASID which has been updated to allow the required EPS message sets

booshibooshi (approved):

PR #478, opened 2021-05-26
Move conversion tests from smoke-tests to unit-tests

booshibooshi:


tests/e2e/pact/resources/test-resources.ts
```diff
@@ -15,7 +15,12 @@ function getTaskCases(operation: ApiOperation) {
     .map(spec => spec.toJestCase())
 }
 
-export const convertCaseGroups = fetcher.convertExamples.filter(e => e.isSuccess).map(spec => spec.toSuccessJestCase())
+// take the first 10 convert success examples as pact struggles with large regex matching
+// todo: we should look to move away from this, investigation into whether TKW validation will handle testing
```
dan-jones-iw, line 19: this begs the question, why are we using convert in the smoke tests? convert checks given FHIR input => expected v3 output. there is no interaction with Spine and validation of the message happens when the same FHIR message is sent to its relevant endpoint. cant convert tests run at build / deploy?

booshibooshi: Good point, we can now load examples in coordinator tests from the models module, I can take a look at comparing input/output for all our examples in unit tests and remove the convert smoke tests (maybe just keep one to prove endpoint is up). It would ensure we're not losing any quality checks and provide feedback faster, bit less visible to SA, but we can look at that in future, maybe TWK validation will help but needs investigation

coordinator/tests/services/translation/convert.successes.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe("conversion tests", () => {
+  test.each(TestResources.convertSuccessExamples)(
+    "should be able to convert %s message to HL7V3",
+    async (_: string, request: unknown, response: string, responseMatcher: string) => {
+      const regex = new RegExp(responseMatcher)
+      const isMatch = regex.test(response)
+      expect(isMatch).toBe(true)
+
+      // copy of convert route logic, todo: either test injecting request into endpoint
```
dan-jones-iw, line 21: creating a test server with just the `/$convert` route seems good :+1:

dan-jones-iw (approved):

PR #479, opened 2021-05-27
Aea 1301 feature toggle on dispense prescribe

dan-jones-iw:


coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -24,6 +24,15 @@ export function verifyBundle(bundle: fhir.Bundle): Array<fhir.OperationOutcomeIs
     return [errors.messageTypeIssue]
   }
 
+  if (process.env.PRESCRIBE_ENABLED !== "true" &&
+      (messageType === fhir.EventCodingCode.PRESCRIPTION || messageType === fhir.EventCodingCode.CANCELLATION)) {
+    return [errors.functionalityDisabled]
```
booshibooshi, line 29: Could we move theses checks to functions in a `feature-toggle.ts` file and import them so we're hiding and reusing logic and this file just has the high-level `if (checkPrescribeFeatureEnabled(messageType))` as an example

booshibooshi (dismissed):

booshibooshi (approved):

PR #483, opened 2021-06-01
AEA-1423: Add translations for patient release requests.

mfjarvis:


coordinator/src/services/translation/request/index.ts
```diff
@@ -154,7 +154,9 @@ export async function convertParametersToSpineRequest(
   logger: pino.Logger
 ): Promise<spine.SpineRequest> {
   const hl7ReleaseRequest = await translateReleaseRequest(parameters, logger)
-  const interactionId = hl7V3.Hl7InteractionIdentifier.NOMINATED_PRESCRIPTION_RELEASE_REQUEST
+  const interactionId = hl7ReleaseRequest instanceof hl7V3.NominatedPrescriptionReleaseRequestWrapper
```
dan-jones-iw, line 157: its a shame we have to check what kind of request it is twice (once here after translation and once before at the start of translations)

mfjarvis: Yeah the whole request translation could do with some significant refactoring

dan-jones-iw (approved):

PR #489, opened 2021-06-08
AEA-1376: Add translations for more Spine errors.

mfjarvis:


coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -135,7 +135,8 @@ export class SpineResponseHandler<T> {
   }
 
   private static toEpsPrescribeErrorCode(code: hl7V3.Code<string>): EpsErrorCodeInformation {
-    switch(code._attributes.code){
+    switch (code._attributes.code) {
+      //TODO - remove?
```
mfjarvis, line 139: The codes with this TODO will never be returned by Spine, but are in the compliance spec. I think we should remove them anyway.

booshibooshi (approved):

PR #517, opened 2021-07-01
Aea 1517 lightweight healthcheck for ecs deploy

callumtipper-IW:


coordinator/src/routes/health/get-status.ts
```diff
@@ -34,28 +59,22 @@ export default [
     path: "/_status",
     handler: async (request: Hapi.Request, h: Hapi.ResponseToolkit): Promise<Hapi.ResponseObject> => {
       const checks: { [name: string]: Array<StatusCheckResponse> } = {
```
mfjarvis, line 61: Maybe inline `checks` here and on line 73, then we won't need the ugly `{ [name: string]: Array<StatusCheckResponse> }` type definition?

coordinator/src/routes/health/get-status.ts
```diff
@@ -11,21 +12,45 @@ export interface StatusCheckResponse {
...
       status: response.status === 200 ? "pass" : "error",
       timeout: "false",
       responseCode: response.status,
       outcome: response.data,
       links: url
-    }))
-    .catch(error => ({
+    }
+  } catch (error) {
+    logger.error(error)
```
mfjarvis, line 26: Might be a good idea to add a log line for some context, e.g. `logger.error(error, "Error calling external service for status check")`

coordinator/src/routes/health/get-status.ts
```diff
@@ -11,21 +12,45 @@ export interface StatusCheckResponse {
...
+    return {
       status: "error",
       timeout: error.code === "ECONNABORTED" ? "true" : "false",
       responseCode: 500
-    }))
+    }
+  }
+}
+
+function createStatusResponse(checks: { [p: string]: Array<StatusCheckResponse> }, h: Hapi.ResponseToolkit) {
```
mfjarvis, line 35: I think the type `{ [p: string]: Array<StatusCheckResponse> }` can also be represented as `Record<String, Array<StatusCheckResponse>>`. Not sure whether that's nicer or not :shrug:

coordinator/src/routes/health/get-status.ts
```diff
@@ -11,21 +12,45 @@ export interface StatusCheckResponse {
...
+  let responseStatus = "pass"
+  let responseCode = 200
+  const warnFilter = Object.values(checks).flat().filter(response => response.status === "warn")
+  if (warnFilter.length > 0) {
```
mfjarvis, line 39: Rather than a filter followed by a length check, you could write this as:
```
const allChecks = Object.values(checks).flat()
if (allChecks.find(check => check.status === "warn")) {
...do warn stuff...
}
if (allChecks.find(check => check.status === "error")) {
...do warn stuff...
}
```

coordinator/src/routes/health/get-status.ts
callumtipper-IW: done

coordinator/src/routes/health/get-status.ts
callumtipper-IW: done

coordinator/src/routes/health/get-status.ts
callumtipper-IW: done

coordinator/src/routes/health/get-status.ts
callumtipper-IW: done

mfjarvis (approved):

PR #518, opened 2021-07-01
AEA-1229: added return task operation to pact test list

callumtipper-IW:


mfjarvis (approved):

coordinator/src/routes/type-guards.ts
```diff
@@ -0,0 +1,31 @@
+import {fhir} from "@models"
```
booshibooshi, line 1: I like this refactor, nicely decoupled from hapi now, but is `routes/` the right place for this now?

booshibooshi (approved):

PR #524, opened 2021-07-06
AEA-1520: Split operations into the individual functions they perform.

mfjarvis:


specification/electronic-prescription-service-api.yaml
```diff
@@ -320,6 +305,64 @@ paths:
...
+      tags:
+        - prescribing
+      parameters:
+        - $ref: '#/components/parameters/BearerAuthorization'
+        - $ref: '#/components/parameters/RequestID'
+        - $ref: '#/components/parameters/CorrelationID'
+      requestBody:
+        required: true
+        content:
+          #FIXME - should be set to application/fhir+json but we have had to change it due to an Apigee SmartDocs bug.
```
dan-jones-iw, line 347: line 316 has this bug 'fixed', wonder if its worth removing these instead of adding more?

specification/electronic-prescription-service-api.yaml
```diff
@@ -348,19 +391,146 @@ paths:
...
+              examples:
+                OperationOutcome:
+                  description: |
+                    An error response.
+                  value:
+                    $ref: components/examples/example-a-validation-error-missing-field/Response-FhirError.json
+  /FHIR/R4/Task#return:
+    post:
+      operationId: return
+      summary: Return a downloaded prescription
```
dan-jones-iw, line 523: is this summary useful for someone reading for the first time?

dan-jones-iw: maybe something along the lines of: 'return a downloaded prescription to ~spine~ so another dispenser can dispense'

dan-jones-iw: rejig the words not the best example

specification/electronic-prescription-service-api.yaml
```diff
@@ -377,6 +547,61 @@ paths:
...
+              examples:
+                example:
+                  description: |
+                    An error response to a return request.
+                  value:
+                    $ref: components/examples/release/not-found-error.json
+  /FHIR/R4/Task#withdraw:
+    post:
+      operationId: withdraw
+      summary: Withdraw a dispense-notification message
```
dan-jones-iw, line 581: maybe get sachin's opinion on these summaries cause i think it can be difficult to sum up what the endpoint does without using overly specific wording?

specification/electronic-prescription-service-api.yaml
mfjarvis: Done

specification/electronic-prescription-service-api.yaml
mfjarvis: Will do

specification/electronic-prescription-service-api.yaml
mfjarvis: I'd prefer to keep the description brief as this appears in the left nav. The first sentence of the description explains the interaction in more detail.

dan-jones-iw (approved):

PR #525, opened 2021-07-07
AEA-1527: Add separate products for prescribing, dispensing and tracker.

mfjarvis:


models/errors/validation-errors.ts
```diff
@@ -8,10 +8,19 @@ export const messageTypeIssue: fhir.OperationOutcomeIssue = {
   expression: ["Bundle.entry.resource.ofType(MessageHeader).eventCoding.code"]
 }
 
-export const featureBlockedIssue: fhir.OperationOutcomeIssue = {
+export const incorrectScopeIssue: fhir.OperationOutcomeIssue = {
   severity: "fatal",
-  code: fhir.IssueCodes.NOT_SUPPORTED,
-  diagnostics: "The feature for this message type is disabled"
+  code: fhir.IssueCodes.FORBIDDEN,
```
mfjarvis, line 13: Wondering if we should return a 403 if the issue code is `FORBIDDEN`? What about `NOT_SUPPORTED` below?

coordinator/src/services/validation/prescribing-dispensing-tracker.ts
```diff
@@ -0,0 +1,39 @@
...
+  "urn:nhsd:apim:user-nhs-id:aal3:electronic-prescription-service-api:dispensing",
+  "urn:nhsd:apim:app:level3:electronic-prescription-service-api:dispensing"
+]
+
+export function validatePermittedPrescribeMessage(scope: string): Array<fhir.OperationOutcomeIssue> {
+  if (!getPrescribeEnabled()) {
+    return [errors.createDisabledFeatureIssue("Prescribing")]
+  }
+
+  if (!validateScope(scope, prescribingPermittedScopes)) {
```
mfjarvis, line 18: If the scopes from the request include the prescribing _app restricted_ scope, we could return a more specific error message here: "Prescribing is not available using the app-restricted access mode."

models/errors/validation-errors.ts
mfjarvis: Made this change. Left `NOT_SUPPORTED` alone.

coordinator/src/services/validation/prescribing-dispensing-tracker.ts
mfjarvis: Made this change.

coordinator/src/utils/status-code.ts
```diff
@@ -0,0 +1,11 @@
+import {fhir} from "@models"
+
+export function getStatusCode(issues: Array<fhir.OperationOutcomeIssue>): number {
+  if (issues.every(i => i.code === fhir.IssueCodes.INFORMATIONAL)) {
+    return 200
+  }
+  if (issues.every(i => i.code === fhir.IssueCodes.FORBIDDEN)) {
+    return 403
+  }
+  return 400
```
ri-ch, line 10: It would be useful to add a log message here so we can see which IssueCodes were not explicitly handling

ri-ch (commented):
It would be useful to have scope validation in the proxy aswell (for all valid scopes). This would be a useful fallback if we ever break the logic in the application.

coordinator/src/utils/status-code.ts
mfjarvis: It will be nearly all of them, but intentionally so

mfjarvis: Will we do anything with this logging?

mfjarvis:
> It would be useful to have scope validation in the proxy aswell (for all valid scopes). This would be a useful fallback if we ever break the logic in the application.

Sure, will add that too :+1:

coordinator/src/utils/status-code.ts
dan-jones-iw: not sure this would really do anything since we do handle all codes explicitly. this is just letting us determine which status code to use for a certain error type

dan-jones-iw (dismissed):

dan-jones-iw (approved):

PR #526, opened 2021-07-08
AEA-1471: Update dispenser notes translation.

mfjarvis:


coordinator/src/services/translation/response/release/release-medication-request.ts
```diff
@@ -187,17 +185,20 @@ export function createSnomedCodeableConcept(code: hl7V3.SnomedCode): fhir.Codeab
...
-    text: dosageInstructions.value._text
+export function createNote(additionalInstructions: string): Array<fhir.Annotation> {
+  if (!additionalInstructions) {
+    return undefined
   }
-  if (additionalInstructions) {
-    dosage.patientInstruction = additionalInstructions
+
+  return [{
+    text: additionalInstructions
```
dan-jones-iw, line 194: will spine ever return multiple patient instructions? (not sure how this would be represented in v3)

dan-jones-iw (approved):

coordinator/src/services/translation/response/release/release-medication-request.ts
mfjarvis: No, there's only a single HL7 V3 field for these

PR #534, opened 2021-07-21
Add patient release to sandbox

booshibooshi:
…acy release

### DIFF
--- a/coordinator/src/services/communication/sandbox-spine-client.ts
+++ b/coordinator/src/services/communication/sandbox-spine-client.ts
@@ -20,6 +20,11 @@ export class SandboxSpineClient implements SpineClient {
           statusCode: 200,
           body: spineResponses.NOMINATED_PRESCRIPTION_RELEASE
         })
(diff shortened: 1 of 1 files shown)


coordinator/src/services/communication/sandbox-spine-client.ts
```diff
@@ -20,6 +20,11 @@ export class SandboxSpineClient implements SpineClient {
           statusCode: 200,
           body: spineResponses.NOMINATED_PRESCRIPTION_RELEASE
         })
+      case hl7V3.Hl7InteractionIdentifier.PATIENT_PRESCRIPTION_RELEASE_REQUEST._attributes.extension:
+        return Promise.resolve({
+          statusCode: 200,
+          body: spineResponses.NOMINATED_PRESCRIPTION_RELEASE
```
mfjarvis, line 26: This will do the job. Might be nice to have a nominated example with 25 prescriptions and a patient example with a single prescription though?

mfjarvis (approved):

coordinator/src/services/communication/sandbox-spine-client.ts
booshibooshi: agreed, would like to get this in quickly and then improve our examples in next pass

PR #536, opened 2021-07-22
implement int preview

dan-jones-iw:
…pdate spec to detail changes

### DIFF
--- a/azure/azure-release-pipeline.yml
+++ b/azure/azure-release-pipeline.yml
@@ -59,6 +59,18 @@ extends:
         depends_on:
           - internal_dev
           - internal_dev_sandbox
+      - environment: int
+        stage_name: int-preview
+        make_spec_visible: true
+        fully_qualified_service_name: electronic-prescription-service-api-int-preview
+        service_base_path: electronic-prescriptions-preview
+#        service_name: electronic-prescriptions-int-preview
+#        short_service_name: epip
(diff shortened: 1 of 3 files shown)


azure/azure-release-pipeline.yml
```diff
@@ -59,6 +59,18 @@ extends:
...
+        make_spec_visible: true
+        fully_qualified_service_name: electronic-prescription-service-api-int-preview
+        service_base_path: electronic-prescriptions-preview
+        service_name: electronic-prescriptions-int-preview
+        short_service_name: epip
+        product_display_name: electronic-prescription-service-api-int-preview
+        proxy_path: live
+        depends_on:
+            - internal_dev
+            - internal_dev_sandbox
```
booshibooshi, line 73: `internal_qa(-sandbox)` depends on `internal_dev(-sandbox)`, I think either this should be updated to depend on `internal_qa(-sandbox)` or we should remove `internal_qa(-sandbox)` from pipeline instead and leave this how it is if we're all agreed these environments are not adding any value for anybody

dan-jones-iw: i agree with removing internal qa but dont want to do that in this pr. feel like there will be a lot of bureaucracy involved if we remove qa without proper sign off and no reason why we cant add preview now and remove qa later?

booshibooshi (approved):

PR #542, opened 2021-07-26
AEA-1552: Added authoredOn and providedBy fields to the release response, tests…

callumtipper-IW:
… for translateAndAddAgentPerson still needed


dan-jones-iw (approved):
code looks good, could we make the merge commit message to master include what new rules the validator change introduces?

dan-jones-iw (approved):

PR #549, opened 2021-08-09
AEA-1555: Fixes to translations for professional & prescribing codes.

mfjarvis:
- Stop modifying professional codes within the API.
- Put spurious code in PractitionerRole instead of Practitioner.
- Where the responsible party is the same person as the author, add details to the author resources instead of creating new ones.
- Similarly where the performer is the same person as the author or responsible party.


coordinator/src/services/translation/response/identifiers.ts
```diff
@@ -0,0 +1,56 @@
...
+  return GPHC_CODE_MATCHER.test(userId)
+}
+export function looksLikeHcpcCode(userId: string): boolean {
+  return HCPC_CODE_MATCHER.test(userId)
+}
+export function looksLikeDinCode(userId: string): boolean {
+  return DIN_CODE_MATCHER.test(userId)
+}
+
+export function createPractitionerOrRoleIdentifier(userId: string): fhir.Identifier {
```
ri-ch, line 33: What you have is pretty explicit and is probably the better way to do it, but if we don't individually need the `looksLike...` functions, you could map the id and its associated schema together, which would avoid the chain of if statements:

```
identifiers = [
	{
		regex: /^G[67][0-9]{6}$/,
		schema: "https://fhir.hl7.org.uk/Id/nhsbsa-spurious-code"
	}
]

createPractitionerOrRoleIdentifier = (userId: string) => {
	for (const id in identifiers) {
		if (id.regex.test(userId)) {
			return fhir.createIdentifier(id.schema, userId)
		}
	}

	return fhir.createIdentifier("https://fhir.hl7.org.uk/Id/professional-code", userId)
}
```

mfjarvis: Yeah I initially thought we'd need at least some of the individual methods, but that turned out not to be the case. Happy to change if you want

mfjarvis: Done

ri-ch (approved):

PR #554, opened 2021-08-23
AEA-1630: Fix the order of pertinent information within a line item.

mfjarvis:


coordinator/src/services/translation/request/line-item.ts
```diff
@@ -38,65 +38,97 @@ function convertDosageInstructions(dosages: Array<fhir.Dosage>, logger: pino.Log
...
+  }
+
+  const controlledDrugWordsExtension = getExtensionForUrlOrNull(
     controlledDrugExtension.extension,
     "quantityWords",
-    "MedicationRequest.extension.extension"
-  ) as fhir.StringExtension : null
-  const controlledDrugWords = controlledDrugWordsExtension?.valueString
-  const controlledDrugWordsWithPrefix = controlledDrugWords ? `CD: ${controlledDrugWords}` : ""
+    "MedicationRequest.extension(\"https://fhir.nhs.uk/StructureDefinition/Extension-DM-ControlledDrug\").extension"
```
dan-jones-iw, line 99: dont have to escape quotes if you use double `'`

dan-jones-iw (approved):

PR #555, opened 2021-08-23
AEA-1626 Update spine url for internal-qa

booshibooshi:


Makefile
```diff
@@ -32,9 +32,8 @@ release:
 	for env in internal-dev-sandbox internal-qa-sandbox sandbox; do \
 		cat ecs-proxies-deploy.yml | sed -e 's/{{ SPINE_ENV }}/veit07/g' | sed -e 's/{{ SANDBOX_MODE_ENABLED }}/1/g' > dist/ecs-deploy-$$env.yml; \
```
mfjarvis, line 33: I know internal-qa-sandbox doesn't actually connect to Spine, but maybe make the same change here just for consistency?

mfjarvis: Actually never mind, we don't do it for sandbox either

mfjarvis (approved):

PR #565, opened 2021-09-03
AEA-1644: Add response fields to GET /Task interaction in sandbox.

mfjarvis:


models/fhir/task.ts
```diff
@@ -4,29 +4,39 @@ import * as patient from "./patient"
 import * as practitionerRole from "./practitioner-role"
 import * as medicationRequest from "./medication-request"
 import * as extension from "./extension"
+import * as medicationDispense from "./medication-dispense"
+import {DispensingReleaseInformationExtension} from "./medication-request"
```
dan-jones-iw, line 8: import style

dan-jones-iw (approved):

PR #567, opened 2021-09-06
AEA-1644: Extend GET /Task sandbox interaction.

mfjarvis:


booshibooshi (approved):

models/fhir/extension.ts
```diff
@@ -48,3 +48,10 @@ export interface ExtensionExtension<T extends Extension> extends Extension {
 export interface BooleanExtension extends Extension {
   valueBoolean: boolean
 }
+
+export type RepeatInformationExtension = ExtensionExtension<UnsignedIntExtension | DateTimeExtension>
```
dan-jones-iw, line 52: this is actually way more readable in code, maybe all our extensions should be like this

dan-jones-iw: for example, on the MedicationRequestOutcome we have: 
```extension: Array<extension.ReferenceExtension<practitionerRole.PractitionerRole>
    | extension.PrescriptionStatusHistoryExtension>
```
its easy to tell from this that it has the status history extension on it but not a clue what the other extension is

models/fhir/extension.ts
```diff
@@ -48,3 +48,10 @@ export interface ExtensionExtension<T extends Extension> extends Extension {
...
+export type ControlledDrugExtension = ExtensionExtension<StringExtension | CodingExtension>
+export type PrescriptionStatusHistoryExtension = ExtensionExtension<CodingExtension | DateTimeExtension>
+export type DispensingInformationExtension = ExtensionExtension<CodingExtension | DateExtension>
+export type DispensingReleaseInformationExtension = ExtensionExtension<DateExtension>
+export type PrescriptionExtension = ExtensionExtension<CodingExtension>
```
dan-jones-iw, line 57: could this be renamed to courseOfTherapyTypeExtension?

PR #572, opened 2021-09-17
remove auth from metadata

dan-jones-iw:


proxies/live/apiproxy/targets/apim.xml
```diff
@@ -1,6 +1,9 @@
 <TargetEndpoint name="apim">
   <PreFlow/>
   <Flows>
+    <Flow name="No-Auth">
```
ri-ch, line 4: I think your naming pattern is a nicer way to do this, so do you want to update the status flow underneath to something like `APIKeyAuthenticated`?

ri-ch (approved):

proxies/live/apiproxy/targets/apim.xml
dan-jones-iw: i think the names below were copied from pds, happy to change

booshibooshi (approved):

PR #580, opened 2021-09-28
adding a function to clear out exisitng prescriptions before running …

callumtipper-IW:
…pact tests


mfjarvis:
Did you mean to add 100+ files to the repo? If not I wonder if we need to change the gitignore?

tests/e2e/pact/broker/verify.ts
```diff
@@ -123,7 +129,53 @@ async function verifyTracker(): Promise<void> {
...
+  nominatedReleases.forEach(async release => {
+    let response
+    do {
+      console.log(
+        "Clearing Prescriptions For: ",
+        getIdentifierParameterByName(release.parameter, "owner").valueIdentifier.value
+      )
+      response = await nominatedRelease(releaseUrl, release)
+    }
+    while ((response && response.data.resourceType !== "OperationOutcome") && !isSandbox)
```
mfjarvis, line 150: Feels a bit weird to have the sandbox condition here. Could we not just exit right at the start of the method if we're in sandbox?

tests/e2e/pact/broker/verify.ts
```diff
@@ -123,7 +129,54 @@ async function verifyTracker(): Promise<void> {
...
+    release,
+    {
+      headers: {
+        "Content-Type": "application/fhir+json; fhirVersion=4.0",
+        "X-Request-ID": uuid.v4(),
+        "X-Correlation-ID": uuid.v4(),
+        "Authorization": `Bearer ${process.env.APIGEE_ACCESS_TOKEN}`
+      }
+    }
+  ).catch(e => console.log(e))
```
mfjarvis, line 167: Might be a good idea not to catch the exception here but instead put a try catch around the do while loop. That way we won't get stuck retrying forever if we get an error response from the API

tests/e2e/pact/broker/verify.ts
```diff
@@ -123,7 +129,54 @@ async function verifyTracker(): Promise<void> {
...
+  nominatedReleases.forEach(async release => {
+    let response
+    do {
+      console.log(
+        "Clearing Prescriptions For: ",
+        getIdentifierParameterByName(release.parameter, "owner").valueIdentifier.value
+      )
+      response = await nominatedRelease(releaseUrl, release)
+    }
+    while (response && response.data.resourceType !== "OperationOutcome")
```
mfjarvis, line 151: Might be nice to make this condition more specific. We could look for the "no more prescriptions available" issue code.

mfjarvis (approved):

mfjarvis (approved):

tests/e2e/pact/broker/verify.ts
mfjarvis: Listening to me is generally a bad idea

PR #626, opened 2021-11-02
Make 'requester' a reference instead of an identifier

ri-ch:


mfjarvis (approved):

mfjarvis (dismissed):

models/fhir/communication-request.ts
```diff
@@ -1,3 +1,4 @@
+import { fhir } from ".."
```
mfjarvis, line 1: Does this work? Feels like a circular reference

dan-jones-iw (approved):

PR #633, opened 2021-11-05
Replace rivets send page with react send page

booshibooshi:


tool/site/client-react/src/pages/sendPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  return <>
+    <Label isPageHeading>Error</Label>
+    <ErrorMessage>An unknown error occurred.</ErrorMessage>
+  </>
+}
+
+interface SendResult {
+  prescription_ids: string[]
```
mfjarvis, line 97: Will this page fail to render for a bulk send? I can't find a valid test pack to check

mfjarvis: Read a bit further and answered my own question - it'll just render the first one. Think that's fine for now

booshibooshi: Yeah bulk send not supported for the moment, put in a placeholder, but will probably change when we decide how to handle pagination in new client

mfjarvis (approved):

tool/site/server_hapi/src/routes/prescribe/send.ts
```diff
@@ -45,15 +50,17 @@ export default [
...
           prescription_ids: prescriptionIds,
           prescription_id: prescriptionIds[0],
           success: true,
           request_xml: sendRequestHl7,
           request: sendRequest,
           response: sendResponse.fhirResponse,
           response_xml: sendResponse.spineResponse
-        }).code(200)
+        }
+        setSessionValue(`signature_token_${signatureToken}`, sendResult, request)
```
mfjarvis, line 62: This key seems a bit odd (we're not storing a signature token here). Session storage as a whole needs some thought, so not overly worried about changing it right now

PR #646, opened 2021-11-16
POC - Allow editing of repeat information in EPSAT

booshibooshi:


tool/site/client-react/src/components/prescription-summary/prescriptionLevelDetails.tsx
```diff
@@ -82,7 +85,14 @@ const PrescriptionLevelDetails: FC<PrescriptionLevelDetailsProps> = ({
       {repeatsIssued &&
       <SummaryList.Row>
         <SummaryList.Key>Issue Number</SummaryList.Key>
-        <SummaryList.Value>{repeatsIssued} of {repeatsAllowed}</SummaryList.Value>
+        {editModeField === "Issue Number"
```
mfjarvis, line 88: Is this a temporary change or a proposed approach to editing as a whole? Having fields that change on mouse enter seems like a really bad idea for accessibility. Is there a problem with the design we agreed on where we have a form followed by a read-only confirmation page?

If we really want the inline editing thing, I'd suggest using a link within a `SummaryList.Actions` component to toggle between viewing and editing instead of the mouse. See https://nhsdigital.github.io/nhsuk-react-components/iframe.html?id=components-summarylist--standard&viewMode=docs

booshibooshi: Agreed

PR #649, opened 2021-11-16
AEA-1822: Add sorting to prescription summary table.

mfjarvis:


tool/site/client-react/src/components/sorter.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [sortConfig, setSortConfig] = React.useState<SortConfig<T>>(config)
+
+  const sortedItems = React.useMemo(() => {
+    const itemsCopy = [...items]
+    if (sortConfig !== null) {
+      itemsCopy.sort((a, b) => {
+        const sortValueA = a[sortConfig.key]
+        const sortValueB = b[sortConfig.key]
+        const directionValue = sortConfig.ascending ? 1 : -1
+        if (sortValueA > sortValueB) {
```
mfjarvis, line 26: This naive comparison seems to work alright for strings, numbers and dates. Could implement custom sorting functions but don't see the need yet

tool/site/client-react/src/components/sorter.tsx
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return sortConfig.ascending ? <ChevronUpMini/> : <ChevronDownMini/>
+  }
+
+  return {sortedItems, sortBy, getIcon}
+}
+
+const ChevronUpMini = styled(ChevronLeft)`
+  transform: rotate(90deg);
+  width: 1em;
```
mfjarvis, line 58: Don't like setting the size manually, but I'm not good enough at CSS to be able to say "make it the same size as the text please"

mfjarvis: At least using a size in em means the icon should scale with the font size

callumtipper-IW (approved):

PR #652, opened 2021-11-18
Aea 1850 inconsistent ods warnings

dan-jones-iw:


coordinator/src/utils/headers.ts
```diff
@@ -42,3 +43,8 @@ export function getScope(headers: Hapi.Util.Dictionary<string>): string {
   const defaultScope = `${PRESCRIBING_USER_SCOPE} ${DISPENSING_USER_SCOPE} ${TRACKER_USER_SCOPE}`
   return process.env.SANDBOX === "1" ? defaultScope : headers[RequestHeaders.SCOPE]
 }
+
+export function getOdsCode(headers: Hapi.Util.Dictionary<string>): string {
+  const defaultOdsCode = "FER21"
```
dan-jones-iw, line 48: probs should have a better default

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -253,6 +256,16 @@ export function verifyDispenseBundle(bundle: fhir.Bundle): Array<fhir.OperationO
     )
   }
 
+  const organizations = actors.filter(actor => actor.type === "Organization")
```
mfjarvis, line 259: Could use `find()` if we only ever expect 0 or 1 result

coordinator/src/services/validation/claim-validator.ts
```diff
@@ -1,15 +1,27 @@
...
 import {validatePermittedDispenseMessage} from "./scope-validator"
 
-export function verifyClaim(claim: fhir.Claim, scope: string): Array<fhir.OperationOutcomeIssue> {
+export function verifyClaim(
+  claim: fhir.Claim, scope: string, accessTokenOds: string
+): Array<fhir.OperationOutcomeIssue> {
+  const validationErrors = []
   if (claim.resourceType !== "Claim") {
-    return [errors.createResourceTypeIssue("Claim")]
+    validationErrors.push(errors.createResourceTypeIssue("Claim"))
```
mfjarvis, line 9: I think this was returning deliberately because if it's not a Claim, we can't really validate anything else

mfjarvis: Same in other places

coordinator/src/services/validation/task-validator.ts
```diff
@@ -21,6 +21,15 @@ export function verifyTask(task: fhir.Task, scope: string): Array<fhir.Operation
   const statusSpecificErrors = performStatusSpecificValidation(task)
   validationErrors.push(...statusSpecificErrors)
 
+  if (task.requester) {
```
mfjarvis, line 24: One to check with Kevin, but I think requester means prescriber. Task messages (return & withdraw) will come from the dispenser, so this will never match. As far as I know, there's no place in the message for the dispenser details (yet).

proxies/live/apiproxy/targets/apim.xml
```diff
@@ -43,6 +43,13 @@
           <Name>FlowCallout.UserRoleService</Name>
           <Condition>(access-mode Equals "user-restricted")</Condition>
         </Step>
+        <!-- Get ODS code from UserInfo -->
+        <Step>
+          <Name>JavaScript.GetODSFromUserInfo</Name>
```
mfjarvis, line 48: The UserRoleService callout is conditional on access mode being user-restricted, which means these two new steps should probably be too

coordinator/src/services/validation/task-validator.ts
mfjarvis: Actually the right field might be owner?

mfjarvis (dismissed):

coordinator/src/utils/headers.ts
booshibooshi: I see the default headers issue has been fixed, wondering if we want to add some tests to cover these, set `sandbox === "0"` and ensure there are no values set for these?

mfjarvis:
All of this should work as long as dispensing messages always come from organisations and never services

coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -50,7 +52,7 @@ export function verifyBundle(bundle: fhir.Bundle, scope: string): Array<fhir.Ope
       messageTypeSpecificErrors = verifyCancellationBundle(bundle)
       break
     case fhir.EventCodingCode.DISPENSE:
-      messageTypeSpecificErrors = verifyDispenseBundle(bundle)
+      messageTypeSpecificErrors = verifyDispenseBundle(bundle, accessTokenOds)
```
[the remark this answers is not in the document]
dan-jones-iw: might avoid cancel in the hopes it changes format soon

mfjarvis (approved):

PR #655, opened 2021-11-23
AEA-1730 Ensure consistent user journeys for EPSAT react pages

booshibooshi:
* Every react result component now has a back button which reloads the form component which led to the result
* React form components which can be used multiple times now have a back button which goes back to previous page
* Too many requests response now captured when client is making requests to the backend
* Redirects are followed to oauth flow when access-token expires
* Load prescription examples are now deployed with epsat so they will be synced with deployed eps and epsat versions
* Remove code from client which has been migrated to react-client

### DIFF
--- a/tool/site/server_flask/routes.py
+++ b/tool/site/server_flask/routes.py
@@ -130,7 +130,6 @@ def post_change_auth():
 
 
 @app.route(HOME_URL, methods=["GET"])
-@exclude_from_auth()
 def get_home():
     return render_rivets_client("home")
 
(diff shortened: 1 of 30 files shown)


tool/site/server_flask/routes.py
```diff
@@ -157,10 +154,9 @@ def download():
         for index, short_prescription_id in enumerate(short_prescription_ids):
             bundle = hapi_passthrough.get_prescription(short_prescription_id)
             zip_file.writestr(f"prepare_request_{index + 1}.json", json.dumps(bundle, indent=2))
-            # todo: fix 'invalid json' issue
-            # if access_token:
-            #     xml, _status_code = make_eps_api_convert_message_request(access_token, bundle)
-            #     zip_file.writestr(f"prepare_request_{index + 1}.xml", xml)
+            if access_token:
```
callumtipper-IW, line 157: note comment: this may break when testpack json download reinstated

callumtipper-IW (approved):

PR #660, opened 2021-11-26
AEA-1730 Store release responses in EPSAT

booshibooshi:


tool/site/server_hapi/src/routes/prescribe/edit.ts
```diff
@@ -9,7 +10,7 @@ export default [
       const prepareBundles = Array.from(request.payload as any[])
       const prescriptionIds: Array<string> = []
       prepareBundles.forEach((prepareBundle: any) => {
-        const prescriptionId = getMedicationRequests(prepareBundle)[0].groupIdentifier.value
+        const prescriptionId = getMedicationRequests(prepareBundle)[0].groupIdentifier?.value ?? ""
```
mfjarvis, line 13: Probably want to skip a prescription if it's not valid rather than adding it to the session

booshibooshi: Added safety handling for failures to get prescriptionId for this and the one in release

mfjarvis (approved):

PR #666, opened 2021-11-30
Add business-status and authored-on params to query API.

mfjarvis:


coordinator/src/routes/tracker/task.ts
```diff
@@ -8,34 +8,66 @@ import {getScope, RequestHeaders} from "../../utils/headers"
...
+  "business-status"?: string
+  "authored-on"?: string | Array<string>
 }
 
-export type ValidQuery = Partial<Record<QueryParam, string>>
-
-interface QueryParamProperties {
+export interface QueryParamDefinition {
+  name: keyof ValidQuery,
+  supportedBySpine: boolean
```
mfjarvis, line 24: Needs a better name. I want something which means "Spine allows queries using this as the only parameter", but the current name suggests something more like "Spine understands this parameter".

coordinator/src/routes/tracker/task.ts
```diff
@@ -74,42 +106,86 @@ export default [{
...
-  const pipeIndex = rawValue?.indexOf("|") || -1
-  if (pipeIndex !== -1) {
-    return rawValue.substring(pipeIndex + 1)
-  }
-  return rawValue
+  throw new Error("Attempting to make tracker request without prescription or patient identifier")
 }
 
 export function filterBundleEntries(bundle: fhir.Bundle, queryParams: ValidQuery): void {
+  console.log(`Got ${bundle.total} results from Spine before filtering.`)
```
mfjarvis, line 134: Remove or use request.logger

coordinator/src/services/validation/query-validator.ts
```diff
@@ -12,43 +15,68 @@ export const validateQueryParameters = (
...
+  return issues
+}
+
+function validateQueryParameter(
+  definition: QueryParamDefinition,
+  values: Array<string>
+): Array<fhir.OperationOutcomeIssue> {
+  const issues: Array<fhir.OperationOutcomeIssue> = []
+
+  if (!definition.isDateParameter && values.length > 1) {
```
mfjarvis, line 48: Not sure if this is actually an error according to the FHIR spec

mfjarvis:
EPSAT changes will be made in another PR because EPSAT PR deploys point at the master branch deploy.

coordinator/src/routes/tracker/task.ts
dan-jones-iw: `validInIsolation`?

coordinator/src/services/validation/query-validator.ts
mfjarvis: Will leave for now

dan-jones-iw (approved):

PR #669, opened 2021-12-03
Aea 1879 release/claim/return/withdraw author details

callumtipper-IW:


coordinator/src/services/translation/request/agent-attended.ts
```diff
@@ -0,0 +1,47 @@
...
+import {createRepresentedOrganization} from "./agent-unattended"
+
+export async function createAuthorForAttendedAccess(
+  professionalCode: string,
+  logger: pino.Logger,
+  organizationCode?: string,
+): Promise<hl7V3.Author> {
+  const agentPerson = await createAgentPersonForAttendedAccess(professionalCode, logger, organizationCode)
+  const author = new hl7V3.Author()
+  author.AgentPerson = agentPerson
```
JackRobinsonIW, line 14: Just thinking a bit, could we add an optional arguement to the `Author` constructor for `AgentPerson` to avoid having to do this line here?

coordinator/src/services/translation/request/agent-attended.ts
```diff
@@ -0,0 +1,47 @@
...
+  organizationCode?: string,
+): Promise<hl7V3.AgentPerson> {
+  const agentPerson = new hl7V3.AgentPerson()
+  agentPerson.id = new hl7V3.UnattendedSdsRoleProfileIdentifier()
+  agentPerson.code = new hl7V3.UnattendedSdsJobRoleCode()
+  const telecom = new hl7V3.Telecom()
+  telecom._attributes = {
+    use: hl7V3.TelecomUse.WORKPLACE,
+    value: "tel:01234567890"
+  }
```
JackRobinsonIW, line 30: Again this might be a project wide choice but if we had optional arguements in a lot of these class constructors we could avoid having to do these manual reassigns

coordinator/src/services/translation/request/dispense/release.ts
```diff
@@ -5,17 +5,22 @@ import {convertMomentToHl7V3DateTime} from "../../common/dateTime"
...
   const prescriptionIdParameter = getIdentifierParameterOrNullByName(fhirReleaseRequest.parameter, "group-identifier")
-  if (prescriptionIdParameter) {
+  if (prescriptionIdParameter && authorParameter) {
     const prescriptionId = prescriptionIdParameter.valueIdentifier.value
-    return await createPatientReleaseRequest(organizationCode, prescriptionId, logger)
+    const authorProfessionalCode = authorParameter.valueIdentifier.value
+    return await createPatientReleaseRequest(organizationCode, authorProfessionalCode, prescriptionId, logger)
   } else {
     return await createNominatedReleaseRequest(organizationCode, logger)
   }
```
JackRobinsonIW, line 26: is this `if` to make typescript happy or is it possible that we could have `(prescriptionIdParameter == truthy && authorParameter == falsey)` or vice versa?

just wondering if this check here could actually end up losing us a parameter by accident?

coordinator/tests/services/translation/request/agent-attended.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  const agentPerson = result.AgentPerson
+  expect(agentPerson.id._attributes.extension).toEqual("999999999999")
+  expect(agentPerson.code._attributes.code).toEqual("R9999")
+  expect(agentPerson.telecom[0]._attributes.value).toEqual("tel:01234567890")
+
+  const agentPersonPerson = agentPerson.agentPerson
+  expect(agentPersonPerson.id._attributes.extension).toEqual("7654321")
+  expect(toArray(agentPersonPerson.name.given)[0]._text).toEqual("Attended")
+  expect(toArray(agentPersonPerson.name.family)[0]._text).toEqual("Access")
+})
```
JackRobinsonIW, line 30: I tend to find unit tests are most useful and maintainable when we have one assertion per test

so here we could either try to condense these expects down to a single one (expect(agentPerson).toStrictEqual(...)) or we could split this test into a describe block i.e

`
describe("user details contain placeholder data - attended") {
    test("the agent person data is correct") {
        expects from lines 22-24 condensed to one assertion
    }
    test("the agent person person data is correct") {
        expects from lines 27-29 condensed to one assertion
    }
`

test names could be better 😅

coordinator/tests/services/translation/request/agent-unattended.spec.ts
```diff
@@ -29,7 +29,7 @@ test("user details contain placeholder data", async () => {
   expect(toArray(agentPersonPerson.name.family)[0]._text).toEqual("Access")
 })
 
-test("organization details are populated from ODS response", async () => {
+test("organization details are populated from ODS response - Unattended", async () => {
```
JackRobinsonIW, line 32: should this be lowercase like the others? `unattended`

coordinator/tests/services/translation/request/dispense/dispense-claim.spec.ts
```diff
@@ -31,13 +31,14 @@ describe("convertDispenseClaim", () => {
 
   test("FHIR replacementOf gets populated in v3", async () => {
     const claim: fhir.Claim = clone(TestResources.examplePrescription3.fhirMessageClaim)
-    claim.extension = [{
+    console.log(claim)
```
JackRobinsonIW, line 34: guessing we don't want to merge this console.log

coordinator/tests/services/translation/request/dispense/release.spec.ts
```diff
@@ -6,37 +6,44 @@ import {
...
+jest.mock("../../../../../src/services/translation/request/agent-attended", () => ({
+  createAuthorForAttendedAccess: jest.fn()
+}))
+
 describe("translateReleaseRequest", () => {
   const mockAuthorResponse = new hl7V3.Author()
   mockAuthorResponse.AgentPerson = new hl7V3.AgentPerson()
-  const mockAuthorFunction = createAuthorForUnattendedAccess as jest.Mock
+  const mockUnattendedAuthorFunction = createAuthorForUnattendedAccess as jest.Mock
+  const mockAttendedAuthorFunction = createAuthorForAttendedAccess as jest.Mock
```
JackRobinsonIW, line 25: think it'd be good to see a beforeEach that clears all mocks here, I think we're good as is but anywhere we are setting/resetting mocks it's good to make sure we are clearing them too

coordinator/src/services/translation/request/agent-attended.ts
dan-jones-iw: i like the clarity it provides tbh, putting a field in a constructor hides which part of the class is being populated

coordinator/src/services/translation/request/agent-unattended.ts
```diff
@@ -55,7 +55,7 @@ async function createRepresentedOrganization(
   return convertOrganization(organization)
 }
 
-function convertOrganization(organization: fhir.Organization): hl7V3.Organization {
+export function convertOrganization(organization: fhir.Organization): hl7V3.Organization {
```
dan-jones-iw, line 58: does this need exporting?

coordinator/src/services/translation/request/agent-unattended.ts
```diff
@@ -41,7 +41,7 @@ function createAgentPersonPersonForUnattendedAccess(): hl7V3.AgentPersonPerson {
   return agentPerson
 }
 
-async function createRepresentedOrganization(
+export async function createRepresentedOrganization(
```
dan-jones-iw, line 44: this isn't necessarily code that is agent-unattended, can we move this to a common file?

dan-jones-iw: a lot of the code is duplicate, we could potentially have a `createAuthor` method that can optionally have a professional code? then all of the woolly outside waffle gets reduced into just the one function

dan-jones-iw: however, in the longer term when we have more than just the professional code passed in to the message, i expect we will be receiving a practitionerRole resource here instead. either way can reduce duplication

coordinator/src/services/translation/request/agent-attended.ts
```diff
@@ -0,0 +1,47 @@
...
+  agentPerson.telecom = [telecom]
+  agentPerson.agentPerson = createAgentPersonPersonForAttendedAccess(professionalCode)
+  if (organizationCode) {
+    agentPerson.representedOrganization = await createRepresentedOrganization(organizationCode, logger)
+  }
+  return agentPerson
+}
+
+export function createAgentPersonPersonForAttendedAccess(professionalCode: string): hl7V3.AgentPersonPerson {
+  const agentPerson = new hl7V3.AgentPersonPerson(new hl7V3.ProfessionalCode(professionalCode))
```
dan-jones-iw, line 40: not sure we want to populate the v3 with professional code. i was expecting the userId from the auth token here? will have to verify with kevin / rob tomorrow if they want professional code in v3 or not

models/hl7-v3/dispense-claim.ts
```diff
@@ -78,6 +79,7 @@ export class DispenseClaimSupplyHeader {
   code: codes.SnomedCode
   effectiveTime: core.Null
   repeatNumber?: core.Interval<core.NumericValue>
+  author: Author
```
mfjarvis, line 82: DispenseClaim doesn't have an `author` field, it has a `legalAuthenticator` field.

coordinator/src/services/translation/request/task.ts
```diff
@@ -1,18 +1,24 @@
 import {fhir, hl7V3} from "@models"
-import {getIdentifierValueForSystem} from "../common"
+import {getExtensionForUrl, getIdentifierValueForSystem} from "../common"
 import * as pino from "pino"
-import {createAuthorForUnattendedAccess} from "./agent-unattended"
+import {createAuthorForAttendedAccess} from "./agent-attended"
 
-export async function createAuthorFromTaskOwnerIdentifier(
-  identifier: fhir.Identifier,
+export async function createAuthorFromProvenanceAgentExtension(
```
dan-jones-iw, line 6: name's a bit confusing here, it takes a task as input and uses two parts of it. maybe rename to `createAuthorFromTask`?

coordinator/src/services/translation/request/withdraw/withdraw.ts
```diff
@@ -29,10 +34,19 @@ export function createRecordTarget(identifier: fhir.Identifier): hl7V3.RecordTar
   return new hl7V3.RecordTargetReference(patient)
 }
 
-export function createAuthor(): hl7V3.AuthorPersonSds {
+export function createAuthor(
```
dan-jones-iw, line 37: can we reuse `createAuthorFromProvenanceAgentExtension`?

coordinator/tests/services/translation/request/agent-attended.spec.ts
dan-jones-iw: that describe block is looking good :+1:

coordinator/tests/services/translation/request/agent-unattended.spec.ts
dan-jones-iw: since we're in the `agent-unattended` file, im not convinced we need to put `- unattended` at the end of every test

dan-jones-iw (approved):

coordinator/src/services/translation/request/dispense/release.ts
```diff
@@ -5,17 +5,22 @@ import {convertMomentToHl7V3DateTime} from "../../common/dateTime"
...
   logger: pino.Logger
 ): Promise<hl7V3.NominatedPrescriptionReleaseRequestWrapper | hl7V3.PatientPrescriptionReleaseRequestWrapper> {
   const organizationParameter = getIdentifierParameterByName(fhirReleaseRequest.parameter, "owner")
   const organizationCode = organizationParameter.valueIdentifier.value
+
+  const authorParameter = getIdentifierParameterOrNullByName(fhirReleaseRequest.parameter, "agent")
+
   const prescriptionIdParameter = getIdentifierParameterOrNullByName(fhirReleaseRequest.parameter, "group-identifier")
-  if (prescriptionIdParameter) {
+  if (prescriptionIdParameter && authorParameter) {
```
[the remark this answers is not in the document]
callumtipper-IW: Yeah it's not correct, I clarified with Matt around if nominated is ALWAYS unattended, it can be both so will need to change this bit

PR #677, opened 2021-12-12
AEA-1955: Publish test report

booshibooshi:


azure/azure-build-pipeline.yml
```diff
@@ -32,3 +32,12 @@ extends:
       - bash: "make test"
         displayName: Run Quality Checks
         workingDirectory: "${{ variables.service_name }}"
+      - task: PublishTestResults@2
```
mfjarvis, line 35: Will this step run if the previous step fails?

mfjarvis (approved):

PR #685, opened 2021-12-21
read telecom from fhir request and added to release translation

JackRobinsonIW:


coordinator/src/services/translation/request/dispense/release.ts
```diff
@@ -39,9 +39,15 @@ export async function createNominatedReleaseRequest(
   const hl7Id = new hl7V3.GlobalIdentifier(uuid.v4())
   const timestamp = convertMomentToHl7V3DateTime(moment.utc())
   const hl7Release = new hl7V3.NominatedPrescriptionReleaseRequest(hl7Id, timestamp)
+  const practitonerRoleTelecom = practitionerRole.telecom[0].value
```
mfjarvis, line 42: Typo in practitioner

mfjarvis: Could pass just the telecom value in rather than the entire PractitionerRole. Might make testing simpler?

mfjarvis: Is there any validation to ensure that a telecom is present? Might null pointer here if not

mfjarvis (approved):

PR #691, opened 2022-01-06
unattended access in epsat

dan-jones-iw:


tool/site/client-react/src/index.tsx
```diff
@@ -28,6 +29,9 @@ async function startApplication(baseUrl: string): Promise<void> {
         <PageContainer>
           <BrowserRouter>
             <Switch>
+              <Route path={`${baseUrl}change-auth`}>
+                <ChangeAuthPage/>
+              </Route>
```
booshibooshi, line 34: The original flow redirected to a default oauth login and once you were logged in the `change-auth` concept made sense, think it's lost meaning now that the redirect to change-auth is the default flow on initial visit to the site, `login` probably makes more sense than `change-auth` in the new context of the standard login flow

dan-jones-iw: yea we thought people would want to test the difference in functionality between user and system level access. so they will likely want to change-auth while already logged in. happy to change to login just would imply you should log out before logging in again (not really thought about log-out at all)

tests/e2e/pact/services/get-access-token.ts
```diff
@@ -1,7 +1,7 @@
 import {createJWT} from "./jwt"
 import axios from "axios"
 import {URLSearchParams} from "url"
-import * as fs from "fs"
+import fs from "fs"
 import path from "path"
 
 /* TODO - get from ADO / env variables */
```
booshibooshi, line 7: Do we want to pickup this todo as part of this pr, or next?

tool/site/client-react/src/index.tsx
booshibooshi: yeah might be easier to make decisions once this is in int. future consideration maybe 👍

tool/site/client-react/src/pages/changeAuthPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const response = await axiosInstance.post<AuthResponse>(
+    `${baseUrl}change-auth`,
+    {authMethod}
+  )
+  redirect(`${response.data.redirectUri}`)
+}
+
+const makeUnattendedLoginRequest = async (baseUrl: string) => {
+  const response = await axiosInstance.post<AuthResponse>(`${baseUrl}unattended-login`)
+
```
booshibooshi, line 45: newline 😉

tool/site/client-react/src/pages/changeAuthPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const {baseUrl} = useContext(AppContext)
+
+  const [pressedUserButton, setPressedUserButton] = useState(false)
+
+  return <>
+    <Label isPageHeading>Change auth</Label>
+    {!pressedUserButton ? <>
+      <Label>Select access level:</Label>
+      <ButtonList>
+        <Button type="submit" onClick={() => setPressedUserButton(true)}>User</Button>
```
booshibooshi, line 18: maybe a rename to help context here? setPressedUserButton -> attendedAccessSelected

tool/site/server_hapi/src/routes/auth/get-unattended-access-token.ts
```diff
@@ -0,0 +1,60 @@
...
+      return responseToolkit.response({e}).code(500)
+    }
+  }
+}
+
+interface TokenResponse {
+  access_token: string
+  expires_in: string
+  token_type: "Bearer"
+}
```
booshibooshi, line 60: 👍

tests/e2e/pact/services/get-access-token.ts
dan-jones-iw: ideally neither lol its work that won't really help us with epsat so would rather not spend time on it. this is just a script to print a token to console, only works for internal-dev atm

tool/site/client-react/src/pages/changeAuthPage.tsx
dan-jones-iw: yea seen this pattern elsewhere, will pick up

booshibooshi (approved):

PR #693, opened 2022-01-13
Add my-prescriptions page to epsat

booshibooshi:


tool/site/client-react/src/pages/homePage.tsx
```diff
@@ -0,0 +1,28 @@
+import {ActionLink, Label} from "nhsuk-react-components"
+import * as React from "react"
+import {useContext} from "react"
```
dan-jones-iw, line 3: duplicate import from line 2

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+            <>
+              <Label isPageHeading>My Prescriptions</Label>
+              <p>You do not have any active prescriptions. Once you create or release a prescription you will see it here</p>
+            </>
+          )
+        }
+        return (
+          <>
+            <Label isPageHeading>My Prescriptions</Label>
+            <PrescriptionGroupTable
```
dan-jones-iw, line 27: thoughts on just having the one table that renders all prescriptions and their current status and decides the actions that a prescription can have based on its status?

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+            />
+          </>
+        )
+      }}
+    </LongRunningTask>
+  )
+}
+
+interface Prescriptions {
+  any: Boolean
```
dan-jones-iw, line 47: think we can infer `prescriptions.any` based on the length of the other two arrays?

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
```diff
@@ -0,0 +1,107 @@
...
+}
+
+interface PrescriptionActionProps {
+  view?: boolean
+  release?: boolean
+  dispense?: boolean
+  claim?: boolean
+}
+
+export const PrescriptionGroupTable: React.FC<PrescriptionGroupTableProps> = ({
```
dan-jones-iw, line 74: wonder if its worth separating this out into its own file?

dan-jones-iw: would be good to decouple the tests for this component from the `MyPrescriptionsPage` component

tool/site/client-react/src/pages/homePage.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {AppContext} from "../index"
+
+const HomePage: React.FC = () => {
+  const {baseUrl} = useContext(AppContext)
+
+  return (
+    <>
```
dan-jones-iw, line 10: do we still want the 'change-auth' / 'login' button at the top?

tool/site/server_hapi/src/services/session.ts
```diff
@@ -25,7 +25,19 @@ export function setSessionValue(key: string, value: unknown, request: Hapi.Reque
...
+
+export function appendToSessionValue(key: string, value: unknown, request: Hapi.Request): void {
+  const existingValue = getSessionValueOrDefault(key, request, [])
+  if (!Array.isArray(existingValue)) {
+    throw Error("Cannot append to session value, session value is not an array")
+  }
+  const mergedValue = existingValue.concat(value)
+  setSessionValue(key, mergedValue, request)
+  if (isLocal()) {
+    console.log(`Saving ${key} to session with value: ${JSON.stringify(mergedValue)}`)
```
dan-jones-iw, line 41: think this line will be logged out twice, once here and once when calling `setSessionValue`

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
booshibooshi: did try that originally, found it harder to use, as when you're performing prescription actions you want to easily see current states of prescriptions grouped together. I'd suggest use it in anger for a while before deciding what you want to change about it

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
booshibooshi: can do, the logic looked horrrible in the client, so i moved it to server, if you prefer it in the client I can change it. let me know if it's a sticking point for you

tool/site/server_hapi/src/services/session.ts
booshibooshi: good spot, removed in next pr

tool/site/client-react/src/pages/myPrescriptionsPage.tsx
booshibooshi: have split out the components from the page

tool/site/client-react/src/pages/homePage.tsx
booshibooshi: good question, i think whilst we're changing our login flow for attended/unattended we hold off making a decision on how we want to handle login flows/ui till this is stable

dan-jones-iw (dismissed):

tool/site/client/index.html
```diff
@@ -131,70 +131,20 @@ <h2 class="nhsuk-error-summary__title" id="error-summary-title">
...
-              <svg class="nhsuk-icon nhsuk-icon__arrow-right-circle" xmlns="http://www.w3.org/2000/svg"
-                viewBox="0 0 24 24" aria-hidden="true">
-                <path d="M0 0h24v24H0z" fill="none"></path>
-                <path
-                  d="M12 2a10 10 0 0 0-9.95 9h11.64L9.74 7.05a1 1 0 0 1 1.41-1.41l5.66 5.65a1 1 0 0 1 0 1.42l-5.66 5.65a1 1 0 0 1-1.41 0 1 1 0 0 1 0-1.41L13.69 13H2.05A10 10 0 1 0 12 2z">
-                </path>
-              </svg>
-              <span class="nhsuk-action-link__text">Validate a FHIR Resource</span>
-            </a>
+            <button class="nhsuk-button" type="submit" onclick="updateAuthMethod('cis2')">CIS 2</button>
```
dan-jones-iw, line 137: why has this been added back? the isLogin stuff should've gone in the previous pr right?

booshibooshi: removed

dan-jones-iw (approved):

PR #696, opened 2022-01-19
AEA-1934 Add prescription return to epsat

booshibooshi:


tool/site/client-react/src/components/prescription-tracker/prescriptionSummaryList.tsx
```diff
@@ -76,8 +82,17 @@ export const PrescriptionSummaryList: React.FC<PrescriptionSummaryProps> = ({
...
+          <p>{status}</p>
+          {status === "To Be Dispensed" &&
+            <StyledPrescriptionActions prescriptionId={id} release />
+          }
+          {status === "With Dispenser" &&
+            <StyledPrescriptionActions prescriptionId={id} returnRelease />
+          }
+        </SummaryList.Value>
       </SummaryList.Row>
+
```
dan-jones-iw, line 95: whitespace :):):):)

tool/site/client-react/src/components/prescriptionActions.tsx
```diff
@@ -7,6 +7,7 @@ interface PrescriptionActionsProps {
   prescriptionId: string
   cancel?: boolean
   release?: boolean
+  returnRelease?: boolean
```
dan-jones-iw, line 10: is `returnRelease` potentially a confusing name? whats wrong with just `return`?

tool/site/client-react/src/formatters/dates.ts
```diff
@@ -11,3 +13,7 @@ export function formatMomentAsDate(date: moment.Moment): string {
 export function formatCurrentDate(): string {
   return moment.utc().format("DD-MMM-YYYY")
 }
```
dan-jones-iw, line 15: a

tool/site/client-react/src/pages/releasePage.tsx
```diff
@@ -120,6 +120,12 @@ const ReleasePage: React.FC<ReleasePageProps> = ({
               </StyledTable>
             </>
           }
+          {releaseResult.withDispenser &&
```
dan-jones-iw, line 123: i like this change :+1:

tool/site/client-react/src/pages/returnPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+  }
+}
+
+function getReturnPharmacy(returnFormValues: ReturnFormValues) {
+  return returnFormValues.pharmacy === "custom"
+    ? returnFormValues.customPharmacy
+    : returnFormValues.pharmacy
+}
+
+export const returnReasons = [
```
dan-jones-iw, line 182: would you mind moving this to the valueSet / codeableConcept area of the codebase? would like to keep it consistent with other reference data

tool/site/client-react/src/pages/returnPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+  returnFormValues: ReturnFormValues
+): Promise<ApiResult> {
+  const returnParameters = createReturn(returnFormValues)
+  const returnResponse = await axiosInstance.post<ApiResult>(`${baseUrl}dispense/return`, returnParameters)
+  return getResponseDataIfValid(returnResponse, isApiResult)
+}
+
+function createReturn(returnFormValues: ReturnFormValues): fhir.Task {
+  const identifier = uuid.v4()
+  const bundleIdentifier = identifier
```
dan-jones-iw, line 118: should this be different to `identifier` (`urn:uuid:${identifier}`)?

tool/site/client-react/src/pages/returnPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+  return {
+    resourceType: "Task",
+    id: identifier,
+    identifier: [
+      {
+        system: "https://tools.ietf.org/html/rfc4122",
+        value: bundleIdentifier
+      }
+    ],
+    status: "rejected",
```
dan-jones-iw, line 129: [apparently](https://simplifier.net/guide/NHSDigital-Medicines/Home/FHIRAssets/AllAssets/Profiles/NHSDigital-Task.guide.md#status), status can be one of "rejected", "cancelled" or "failed" for a return, is it worth having that editable in the UI?

tool/site/client-react/src/pages/returnPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+          display: returnReasons.find(r => r.value === returnFormValues.reason).text
+        }
+      ]
+    }
+  }
+}
+
+function getReturnPharmacy(returnFormValues: ReturnFormValues) {
+  return returnFormValues.pharmacy === "custom"
+    ? returnFormValues.customPharmacy
```
dan-jones-iw, line 178: is the value for `customPharmacy` set anywhere? i cant see in the code. cant we just override the value for `pharmacy` in place wherever `customPharmacy` is set?

tool/site/client-react/src/pages/returnPage.tsx
```diff
@@ -0,0 +1,217 @@
...
+import * as uuid from "uuid"
+import {formatCurrentDateTimeIsoFormat} from "../formatters/dates"
+import PharmacyRadios from "../components/pharmacies"
+import RadioField from "../components/radioField"
+
+interface ReturnPageProps {
+  prescriptionId?: string
+}
+
+interface ReturnFormValues {
```
dan-jones-iw, line 24: would you mind moving the returnForm out of the return page? just helps keep files small

tool/site/server_flask/routes.py
```diff
@@ -322,6 +322,21 @@ def post_release():
     return app.make_response(response)
 
 
+@app.route("/dispense/return", methods=["GET"])
+def get_return():
+    if (config.ENVIRONMENT == "prod"):
```
dan-jones-iw, line 327: i know this should never be hit but wondering if that has been added for any particular reason?

tool/site/server_hapi/src/services/communication/live-eps-client.ts
```diff
@@ -52,6 +52,18 @@ export class LiveEpsClient implements EpsClient {
...
 
+  async makeReturnRequest(body: Task): Promise<EpsResponse<OperationOutcome>> {
+    const requestId = uuid.v4()
+    const rawResponseHeaders = {
+      "x-raw-response": "true"
+    }
+    const response = await this.makeApiCall<OperationOutcome>("Task", body, undefined, requestId)
+    const statusCode = response.status
+    const fhirResponse = response.data
+    const spineResponse = (await this.makeApiCall<string | OperationOutcome>("Task", body, undefined, requestId, rawResponseHeaders)).data
```
dan-jones-iw, line 63: another one for another pr, would be good to only do a spineResponse call if we know that the FHIR response has a 200 (to stop validation error happening twice)

tool/site/client-react/tests/pages/__snapshots__/prescriptionSearchPage.spec.tsx.snap
```diff
@@ -3,14 +3,14 @@
 exports[`Clicking back from the detail search results returns to the summary search 1`] = `
 "<h1 class=\\"nhsuk-label-wrapper\\"><label class=\\"nhsuk-label nhsuk-label--xl\\">Search Results</label></h1>
 <div class=\\"nhsuk-table-responsive\\">
-  <table class=\\"nhsuk-table sc-jRQBWg jFVpJk\\">
+  <table class=\\"nhsuk-table sc-gKclnd cgRkdO\\">
```
dan-jones-iw, line 6: not really sure why all these things have changed? dont know if it makes a difference really?

tool/site/client-react/src/components/prescriptionActions.tsx
booshibooshi: return is a reserved word (as in return from function) so won't compile. Yeah definitely worth considering, agree not in this pr though

tool/site/client-react/src/components/prescription-tracker/prescriptionSummaryList.tsx
booshibooshi: removed

tool/site/client-react/src/formatters/dates.ts
booshibooshi: b

tool/site/server_flask/routes.py
booshibooshi: being consistent with previous routes, if we start to consider the use-case for a prod environment then our initial assumption was that this should be read only, so all dispense interactions should be unavailable

tool/site/client-react/src/pages/returnPage.tsx
booshibooshi: you can see this behaviour for the PharmacyRadios component which is also used on release page, it wouldn't be a straightforward change I think, because we would be fighting against the benefits that Formik gives us

tool/site/client-react/tests/pages/__snapshots__/prescriptionSearchPage.spec.tsx.snap
booshibooshi: I wouldn't try to determine the diffs here, if you want to see what's changed you can check out the branch and use jest to see the difference between the snapshots by getting the old ones from master and running the test. Summary: "Return Prescription" prescription action now displays on this page when a prescription is with a dispenser and "Release prescription" see `prescriptionSummaryList` component under `prescription-tracker`

tool/site/client-react/src/pages/returnPage.tsx
booshibooshi: moved

tool/site/client-react/src/pages/returnPage.tsx
booshibooshi: moved

tool/site/client-react/src/pages/returnPage.tsx
booshibooshi: identifiers are hardcoded as they have no impact on the result of the return call, the urn:uuid... convention is used for fullUrl which is not required here

tool/site/client-react/src/pages/returnPage.tsx
booshibooshi: I think generally we should carry on following the approach to add minimum required editable fields until we have requests otherwise. Can always add this in if we get a requirement to in future, but the added value here is being able to return and re-release prescriptions

dan-jones-iw (approved):

PR #698, opened 2022-01-19
use new simplifier version in OAS

dan-jones-iw:


specification/electronic-prescription-service-api.yaml
```diff
@@ -150,7 +150,7 @@ info:
     To begin the onboarding process please download and complete the [EPS use case form](https://nhs-prod.global.ssl.fastly.net/binaries/content/assets/website-assets/developer/electronic-prescription-service-fhir-api/eps-access-form-v1.0-org-short-name-yyyymmdd-1.docx) and email it to us at [epsonboarding@nhs.net](mailto:epsonboarding@nhs.net).
 
     ## Resources
-    Use the [Digital Medicines Implementation Guide](https://simplifier.net/guide/DigitalMedicines) to assist with your integration.
+    Use the [Digital Medicines Implementation Guide](https://simplifier.net/guide/NHSDigital-Medicines) to assist with your integration.
```
JackRobinsonIW, line 153: It'd be nice if we could declare some kind of `baseUrl` variable, to make these changes smaller in the future, don't think there is an easy way to do this with yaml sadly

JackRobinsonIW (approved):

PR #702, opened 2022-01-22
Update auth flows for epsat

booshibooshi:
Ensure int is authenticated by cis2 only and internal-dev does not display unusable cis2 option


tool/site/client-react/src/pages/loginPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+import ButtonList from "../components/buttonList"
+import {redirect} from "../browser/navigation"
+import {isDev, isInt} from "../services/environment"
+
+const LoginPage: React.FC = () => {
+  const {baseUrl, environment} = useContext(AppContext)
+
+  const [attendedAccessSelected, setAttendedAccessSelected] = useState(false)
+
+  if (isInt(environment)) {
```
dan-jones-iw, line 14: this would stop devs without smartcards from using int - is this what we want?

tool/site/client-react/tests/pages/__snapshots__/loginPage.spec.tsx.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Displays user/system options in internal-dev 1`] = `
```
dan-jones-iw, line 3: delete the change-auth snap?

tool/site/client-react/src/pages/loginPage.tsx
booshibooshi: short answer: yes, we should all have smartcards. will discuss in more details

tool/site/client-react/tests/pages/__snapshots__/loginPage.spec.tsx.snap
booshibooshi: there was no change-auth snap as there were no tests covering this area before I added them

dan-jones-iw: lol :clown_face:

callumtipper-IW (approved):

PR #719, opened 2022-01-27
Remove old client from epsat

booshibooshi:


JackRobinsonIW (approved):
lgtm, some large file diffs but couldn't spot anything

PR #730, opened 2022-01-31
Validation hotfix

dan-jones-iw:


JackRobinsonIW (dismissed):

booshibooshi (changes requested):
We shouldn't be reverting all these epsat changes, only the dispensing change

booshibooshi (approved):

PR #740, opened 2022-02-04
prefexed telecom on agent person translations

JackRobinsonIW:


coordinator/src/services/translation/request/agent-unattended.ts
```diff
@@ -106,19 +106,24 @@ export async function createAgentPerson(
...
-  }
 
   const agentPerson = new hl7V3.AgentPerson()
   agentPerson.id = new hl7V3.SdsRoleProfileIdentifier(sdsRoleProfileId)
   agentPerson.code = new hl7V3.SdsJobRoleCode(sdsJobRoleCode)
-  agentPerson.telecom = [v3Telecom]
   agentPerson.agentPerson = createAgentPersonPerson(sdsUserUniqueId, name)
   agentPerson.representedOrganization = representedOrganisation
 
+  // dont think I want this check, 90% sure we should have a telecom value
```
JackRobinsonIW, line 116: Added this check because of failing tests, I think the actual fix is to make sure the tests provide some kind of telecom and then this if should be more like

```
const providedTelecom = fhirTelecom || representedOrganisation.telecom?._attributes.value
if (!!providedTelecom) {
  throw new Error("Telecom missing blah blah blah...")
}
```

dan-jones-iw (approved):

PR #742, opened 2022-02-08
withdraw removes dispense notification from session storage

JackRobinsonIW:


tool/site/server/src/routes/dispense/withdraw.ts
```diff
@@ -14,6 +20,22 @@ export default [
       const withdrawResponse = await epsClient.makeWithdrawRequest(withdrawRequest)
       const withdrawRequestHl7 = await epsClient.makeConvertRequest(withdrawRequest)
       const success = withdrawResponse.statusCode === 200
+      if (success) {
+        const prescriptionId = withdrawRequest.groupIdentifier?.value
+        const dispenseNotificationRequestKey = `dispense_notification_requests_${prescriptionId}`
+
+        const dispenseNotificationRequests = getSessionValue(dispenseNotificationRequestKey, request) as Bundle[]
```
dan-jones-iw, line 27: `Array<Bundle>` :pray:

tool/site/server/src/routes/dispense/withdraw.ts
```diff
@@ -14,6 +20,22 @@ export default [
...
+        const newDispenseNotifications = dispenseNotificationRequests.slice(0, -1)
```
dan-jones-iw, line 28: ive seen array.pop() used to remove the last element from the array. not sure its useful here since you want all elements except the last.

dan-jones-iw: can this ever be run on an empty array? (i think the -1 would throw?)

dan-jones-iw (approved):

tool/site/server/src/routes/dispense/withdraw.ts
JackRobinsonIW: Done

tool/site/server/src/routes/dispense/withdraw.ts
JackRobinsonIW: yeah so array.pop() removes and returns the last element so could use it but I generally prefer not to mutate state so rather than mutating `dispenseNotifications` I declared a new const `newDispenseNotifications`. Not precious about this though so happy to change if preferred

I did wonder and was fairly sure this shouldn't run on and empty array but even if it does the `.slice(0, -1)` will just return an empty array which is probably the preferred behaviour

tool/site/client/src/pages/withdrawPage.tsx
```diff
@@ -27,36 +27,50 @@ const WithdrawPage: React.FC<WithdrawPageProps> = ({
...
+              </>
+            )}
+          </LongRunningTask>
+        )
+      }}
     </LongRunningTask>
   )
 }
 
+async function retrieveDispenseNotifications(baseUrl: string, prescriptionId: string): Promise<{[key: number]: fhir.Bundle}> {
```
dan-jones-iw, line 68: change the type to `Promise<{dispenseNotifications: Array<fhirBundle>}>`

JackRobinsonIW: Done

callumtipper-IW (approved):

PR #743, opened 2022-02-09
AEA-2042 Disable RSS in EPSAT

booshibooshi:


tool/ecs-proxies-deploy.yml
```diff
@@ -31,12 +31,6 @@ docker_service:
...
             -   name: DEMO_APP_PRIVATE_KEY
                 valueFrom: "arn:aws:secretsmanager:eu-west-2:{{ PTL_ACCOUNT_ID }}:secret:{{ account }}/eps-api-tool/{{ APIGEE_ENVIRONMENT }}/test-client/local-jwt-private-key"
-            -   name: APP_JWT_PRIVATE_KEY
-                valueFrom: "arn:aws:secretsmanager:eu-west-2:{{ PTL_ACCOUNT_ID }}:secret:{{ account }}/eps-api-tool/{{ APIGEE_ENVIRONMENT }}/test-client/rss-jwt-private-key"
-            -   name: DEMO_APP_REMOTE_SIGNING_KID
-                valueFrom: "/{{ account }}/api-deployment/eps-api-tool/{{ APIGEE_ENVIRONMENT }}/test-client/rss-jwt-key-id"
-            -   name: DEMO_APP_REMOTE_SIGNING_ISSUER
-                valueFrom: "/{{ account }}/api-deployment/eps-api-tool/{{ APIGEE_ENVIRONMENT }}/test-client/rss-jwt-issuer"
             -   name: APP_JWT_SUBJECT
                 valueFrom: "/{{ account }}/api-deployment/eps-api-tool/int/test-client/rss-jwt-subject"
```
JackRobinsonIW, line 35: Same question about `rss-jwt-subject` here

tool/site/client/src/pages/configPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const initialValues = {useSigningMock: false, signingPrNumber: undefined}
+
+  useEffect(() => {
+    if (configFormValues) {
+      (async () => {
+        const success = await updateConfig(baseUrl, configFormValues)
+        setConfigUpdateSuccess(success)
+      })()
+    }
+  }, [baseUrl, configFormValues, setConfigUpdateSuccess])
```
JackRobinsonIW, line 28: I'm a little confused why we are using a `useEffect` here? 

If I've understood the logic when we click `Save` that calls `setConfigFormValues` which triggers this useEffect, this does the updated if the `configFormValues` are correct.

Could we not have this logic inside the useEffect being triggered directly by the buttons `onSubmit` handler?

tool/site/client/src/pages/configPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+            </ButtonList>
+          </Form>
+        }
+      </Formik>
+    </>
+  )
+}
+
+async function updateConfig(baseUrl: string, configFormValues: ConfigFormValues): Promise<boolean> {
+  return await (await axiosInstance.post(`${baseUrl}config`, configFormValues)).data.success
```
JackRobinsonIW, line 73: do we need this double `await` here?

tool/site/server/src/services/environment.ts
```diff
@@ -1,9 +1,17 @@
-import {CONFIG} from "../config"
+export type Environment = "internal-dev" | "internal-dev-sandbox" | "internal-qa" | "int"
 
-export function isLocal(): boolean {
-  return CONFIG.environment.endsWith("-sandbox") ?? false
+export const internalDev = "internal-dev"
+export const internalDevSandbox = "internal-dev-sandbox"
+export const int = "int"
```
JackRobinsonIW, line 5: yeah I think this might be easier to manage as an enum

tool/ecs-proxies-deploy-internal-qa.yml
```diff
@@ -43,6 +35,8 @@ docker_service:
                 valueFrom: "arn:aws:secretsmanager:eu-west-2:{{ PTL_ACCOUNT_ID }}:secret:{{ account }}/eps-api-tool/internal-qa/test-client/client-key"
             -   name: DEMO_APP_PRIVATE_KEY
                 valueFrom: "arn:aws:secretsmanager:eu-west-2:{{ PTL_ACCOUNT_ID }}:secret:{{ account }}/eps-api-tool/int/test-client/local-jwt-private-key"
+            -   name: APP_JWT_SUBJECT
+                valueFrom: "/{{ account }}/api-deployment/eps-api-tool/int/test-client/rss-jwt-subject"
```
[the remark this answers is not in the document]
booshibooshi: yeah the config for rss/local signing is in a but of a state, needs resolving in a new pr, this subject is the subject in the JWT and is supposed to be the same for both, subject is required for local signing too. Should clean these up in future PR

tool/ecs-proxies-deploy.yml
booshibooshi: Same as above comment

tool/site/client/src/pages/configPage.tsx
booshibooshi: I've added a type for the response here and removed the wrapped await

tool/site/client/tests/pages/configPage.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {axiosInstance} from "../../src/requests/axiosInstance"
+import {internalDev} from "../../src/services/environment"
+import ConfigPage from "../../src/pages/configPage"
+
+const baseUrl = "baseUrl/"
+const context: AppContextValue = {baseUrl, environment: internalDev}
+
+const configUrl = `${baseUrl}config`
+
+beforeEach(() => moxios.install(axiosInstance))
```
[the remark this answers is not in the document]
booshibooshi: I don't think that's the point of these tests. These tests cover the component behaviour which is why the api calls are stubbed. We're not testing server logic here, that should be done elsewhere. The result of the change to the render of the component is being tested here, and that's captured by 2 snapshots, one before the data is sent, and one after a successful result is sent.

tool/site/server/src/config.ts
```diff
@@ -12,26 +14,20 @@ export interface Config {
...
-  rssIssuer: string
 }
 
 export const CONFIG: Config = {
   commitId: process.env.COMMIT_ID ?? "unknown",
   validatorVersion: process.env.VALIDATOR_VERSION ?? "unknown",
   basePath: process.env.BASE_PATH ?? "eps-api-tool",
   baseUrl: process.env.BASE_PATH ? `/${process.env.BASE_PATH}/` : "/",
-  environment: process.env.ENVIRONMENT ?? "int",
+  environment: process.env.ENVIRONMENT as Environment ?? "int",
```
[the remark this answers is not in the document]
booshibooshi: can look at changing it outside of this pr

tool/site/server/src/services/environment.ts
booshibooshi: same as above comment

tool/site/server/src/config.ts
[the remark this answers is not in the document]
JackRobinsonIW: 👍

JackRobinsonIW (approved):

PR #744, opened 2022-02-10
AEA-2042 Handle signing service redirects in EPSAT

booshibooshi:


tool/site/server/src/services/communication/live-signing-client.ts
```diff
@@ -19,7 +20,10 @@ export class LiveSigningClient implements SigningClient {
 
   async uploadSignatureRequest(prepareResponses: Parameters[]): Promise<any> {
     const baseUrl = this.getBaseUrl()
-    const url = `${baseUrl}/signaturerequest`
+    const stateJson = {prNumber: getPrNumber(CONFIG.basePath)}
+    const stateString = JSON.stringify(stateJson)
+    const state = Buffer.from(stateString, "utf-8").toString("base64")
```
JackRobinsonIW, line 25: again my knowledge gap but why do we need the `base64` parsing

JackRobinsonIW (approved):

tool/site/server/src/services/communication/live-signing-client.ts
booshibooshi: it gets passed to a query string parameter

tool/site/server/src/app.ts
```diff
@@ -139,7 +139,7 @@ function addViewRoutes(server: Hapi.Server) {
   server.route(addView("validate"))
   server.route(addView("search"))
   server.route(addView("prescribe/load"))
-  server.route(addView("prescribe/send"))
+  server.route(addView("prescribe/send", true))
```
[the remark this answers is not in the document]
booshibooshi: endpoint needs to be open as it's a callback from signing service and redirects through feature branches. there are safe guards for this route which check query parameters are valid etc, it matches the behaviour of the oauth callback route

PR #747, opened 2022-02-11
EPSAT change to make full dispense to dispense the remaining amount

callumtipper-IW:


tool/site/client/src/components/dispense/createDispenseNotification.ts
```diff
@@ -170,13 +170,15 @@ function createMedicationDispenseType(lineItemStatus: LineItemStatus): fhir.Code
...
-  {statusCode, priorStatusCode, suppliedQuantityValue}: LineItemFormValues
+  {statusCode, priorStatusCode, suppliedQuantityValue, dispensedQuantityValue, prescribedQuantityValue}: LineItemFormValues
 ): fhir.Quantity {
   const dispensedQuantity = {...requestedQuantity}
   if (statusCode === LineItemStatus.PARTIALLY_DISPENSED) {
     dispensedQuantity.value = parseInt(suppliedQuantityValue)
   } else if (statusCode !== LineItemStatus.DISPENSED || priorStatusCode === LineItemStatus.DISPENSED) {
     dispensedQuantity.value = 0
+  } else if (statusCode === LineItemStatus.DISPENSED && priorStatusCode === LineItemStatus.PARTIALLY_DISPENSED) {
+    dispensedQuantity.value = prescribedQuantityValue - dispensedQuantityValue
```
JackRobinsonIW, line 181: based on the `parseInt` on line 177 in fairly sure `prescribedQuantityValue` and `dispensedQuantityValue` are both going to be strings

I think at runtime js might actually be fine with that and just parse them as Ints by default but we are kinda lying with our typescript here about whats going on so would be nice to wrap them both in `parseInt()`

tool/site/client/src/components/dispense/createDispenseNotification.ts
```diff
@@ -170,13 +170,15 @@ function createMedicationDispenseType(lineItemStatus: LineItemStatus): fhir.Code
...
   }
   return dispensedQuantity
 }
```
JackRobinsonIW, line 184: Not sure what our tests look like in this area right now but would be good to see this specific scenario covered as I think the logic is deceptively complex

tool/site/client/src/components/dispense/createDispenseNotification.ts
callumtipper-IW: non need, they don't come from the form so not strings

tool/site/client/src/components/dispense/createDispenseNotification.ts
callumtipper-IW: done

JackRobinsonIW (approved):

PR #759, opened 2022-02-14
AEA-2046: Adding a different dispensed medication from requested opti…

callumtipper-IW:
…on, only for a predefined medication snomed=39720311000001101


tool/site/client/src/pages/dispensePage.tsx
```diff
@@ -162,6 +162,8 @@ export function createStaticLineItemInfo(
     lineItemInfo.priorStatusCode = getLineItemStatus(latestMedicationDispense)
   }
 
+  lineItemInfo.alternativeMedicationAvailable = containsParacetamol(medicationRequest)
+
```
booshibooshi, line 166: on this page when the dispense result shows, if the prescription is still with dispenser - active we should have a `dispense` prescription action available

booshibooshi (approved):

PR #760, opened 2022-02-15
AEA-2057 Add edit mode to EPSAT prescription summary page 

booshibooshi:


tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -1,29 +1,59 @@
...
 interface SendPreSignPageProps {
   prescriptionId: string
 }
 
+interface SendPreSignPageFormValues {
+  numberOfCopies: string
+  nominatedOds: string
+}
+
+type SendPreSignPageFormErrors = PrescriptionSummaryErrors
```
JackRobinsonIW, line 28: bit weird to just copy over this type, if its just to do the rename we could do that as part of the import I think

`import PrescriptionSummaryErrors as SendPreSignPageFormErrors from ...`

tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -1,29 +1,59 @@
...
   const {baseUrl} = useContext(AppContext)
-  const [sendConfirmed, setSendConfirmed] = useState<boolean>(false)
+  const [editMode, setEditMode] = useState(false)
+  const [sendPageFormValues, setSendPageFormValues] = useState<SendPreSignPageFormValues>()
   const retrievePrescriptionTask = () => retrievePrescription(baseUrl, prescriptionId)
 
+  const validate = (values: SendPreSignPageFormValues) => {
+    const errors: FormikErrors<SendPreSignPageFormErrors> = {}
+
+    const copiesError = "Please provide a number of copies between 1 and 25"
```
JackRobinsonIW, line 41: I assume this 25 is just a soft cap we impose right? no actual limits in the API

tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -1,29 +1,59 @@
...
+    if (!values.numberOfCopies) {
+      errors.numberOfCopies = copiesError
+    } else {
+      const copies = parseInt(values.numberOfCopies)
+      if (isNaN(copies)) {
+        errors.numberOfCopies = copiesError
+      }
+      if (copies > 25) {
+        errors.numberOfCopies = copiesError
+      }
```
JackRobinsonIW, line 51: its the same error in either case so we could just do 
```
      if (isNaN(copies) || copies > 25) {
        errors.numberOfCopies = copiesError
      }
```
I think, this also doesn't cover negative numbers or 0but assuming our users aren't trying to break it I think thats probs fine

tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -53,20 +83,36 @@ const SendPreSignPage: React.FC<SendPreSignPageProps> = ({
   return (
     <LongRunningTask<Bundle> task={retrievePrescriptionTask} loadingMessage="Retrieving prescription details.">
       {bundle => {
-        if (!sendConfirmed) {
-          const summaryViewProps = createSummaryPrescription(bundle)
+        if (!sendPageFormValues) {
```
JackRobinsonIW, line 86: Almost wondering if it might be easier to just always use the edit form and just pre-poulate the form fields with the default values? then we could avoid having to manage edit mode at all?

tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -88,13 +134,33 @@ async function retrievePrescription(baseUrl: string, prescriptionId: string): Pr
...
+    })
+    const newPrescriptions: Array<Bundle> = prescriptions
+      .map(p => Array(parseInt(sendPageFormValues.numberOfCopies)).fill(p).map(p => clone(p))
+      ).flat()
+    newPrescriptions.forEach(p => updateBundleIds(p))
+    await axiosInstance.post(`${baseUrl}prescribe/edit`, newPrescriptions)
+  }
+}
+
+function clone(p: any): any {
```
JackRobinsonIW, line 160: could be worth adding `lodash` into the project, has some nice utils that we could use across the project

tool/site/client/src/pages/sendPreSignPage.tsx
```diff
@@ -88,13 +134,33 @@ async function retrievePrescription(baseUrl: string, prescriptionId: string): Pr
...
+    prescriptions.forEach(prescription => {
+      const medicationRequests = getMedicationRequestResources(prescription)
+      medicationRequests.forEach(medication => medication.dispenseRequest.performer.identifier.value = sendPageFormValues.nominatedOds)
+    })
+    const newPrescriptions: Array<Bundle> = prescriptions
+      .map(p => Array(parseInt(sendPageFormValues.numberOfCopies)).fill(p).map(p => clone(p))
+      ).flat()
+    newPrescriptions.forEach(p => updateBundleIds(p))
+    await axiosInstance.post(`${baseUrl}prescribe/edit`, newPrescriptions)
+  }
```
JackRobinsonIW, line 157: 3 forEachs and 2 maps 🧐 I wonder if we could refactor this to be a little easier to read? A

Also could we move the to level if statement to where is function is called rather than here? Bit weird to call this function which might then fail the check and do nothing

tool/site/client/src/pages/sendPreSignPage.tsx
booshibooshi: interesting, I didn't spot this because the linter wrote this code for me

what i wrote was 

```interface SendPreSignPageFormErrors extends PrescriptionSummaryErrors {}```

and although it looks weird now, it's about having seperation of concerns, the Prescription Summary Component should not know the send pre sign page exists, so the Prescription summary component has its own errors, the page has it's own errors which could be a composite of component errors

tool/site/client/src/pages/sendPreSignPage.tsx
booshibooshi: signing-service limit (according to EPS docs: https://digital.nhs.uk/developer/api-catalogue/electronic-prescription-service-fhir#api-Dispensing-verify-signature)

tool/site/client/src/pages/sendPreSignPage.tsx
booshibooshi: pushed this change and handled negative number case too

tool/site/client/src/pages/sendPreSignPage.tsx
booshibooshi: refactored and removed the if statement which always evaluated to true, since we populate it with a default value

tool/site/client/src/pages/sendPreSignPage.tsx
JackRobinsonIW: yeah the extends version make a little more sense to me but I do see why the linter has done it

tool/site/client/src/pages/sendPreSignPage.tsx
JackRobinsonIW: yeah much better thank you

JackRobinsonIW (approved):

PR #763, opened 2022-02-17
AEA-2035 Acute and Repeat Prescribing Examples in EPSAT pass signature verification

booshibooshi:


coordinator/src/services/translation/response/common.ts
```diff
@@ -221,6 +229,11 @@ export function translateAgentPerson(agentPerson: hl7V3.AgentPerson): Translated
 
       return translatedAgentPerson
     } else {
+      const representedOrganization = agentPerson.representedOrganization
```
JackRobinsonIW, line 232: minor but we could pull out `const representedOrganization = agentPerson.representedOrganization` from lines 198 and 232 to be before the above `if` and then use it on line 197

JackRobinsonIW (approved):

JackRobinsonIW (approved):

PR #769, opened 2022-02-18
AEA-2035 Fix repeat prescribing

booshibooshi:


coordinator/src/services/translation/response/release/release-medication-request.ts
```diff
@@ -31,6 +31,8 @@ export function createMedicationRequest(
     ? fhir.MedicationRequestIntent.REFLEX_ORDER
     : fhir.MedicationRequestIntent.ORDER
   const isReflexOrder = intent === fhir.MedicationRequestIntent.REFLEX_ORDER
+  const isContinuous = courseOfTherapyType === fhir.COURSE_OF_THERAPY_TYPE_CONTINUOUS
+  const hasRepeatsAllowed = parseInt(lineItem.repeatNumber?.high?._attributes.value) - 1 > 0
```
JackRobinsonIW, line 35: just wondering if `lineItem.repeatNumber?.high?._attributes.value` is undefined then `parseInt` returns `NaN` I think, if that happens does this boolean still evaluate as we want it too?

booshibooshi: NaN - 1 evaluates as NaN which evaluates as false when checked if more than 0

callumtipper-IW (approved):

PR #775, opened 2022-02-21
allow qa env to use user auth

dan-jones-iw:


booshibooshi:
Header will need updating to allow route to config to use mock signing, see `pageContainer.tsx`

booshibooshi (changes requested):
Header will need updating to allow route to config to use mock signing, see `pageContainer.tsx` `isDev` condition

PR #783, opened 2022-02-23
Fix session timer

booshibooshi:


tool/site/client/src/components/sessionTimer.tsx
```diff
@@ -72,4 +72,8 @@ export const SessionTimer: React.FC = () => {
   )
 }
 
+function getUtcEpochSeconds(date: Date) {
+  return (date.getTime() + date.getTimezoneOffset() * 60 * 1000) / 1000
```
JackRobinsonIW, line 76: I'm not sure we need this logic here, I think the `date.getTime() + date.getTimezoneOffset() * 60 * 1000` could just be replaced by `Date.now()` I think `getTime()` is the problem because its localized but `now()` should just give a unix timestamp

JackRobinsonIW (approved):

PR #784, opened 2022-02-24
AEA-2094: Pact tests for dispense notification amend

callumtipper-IW:


tests/e2e/pact/resources/common.ts
```diff
@@ -5,7 +5,8 @@ export const basePath = "/FHIR/R4"
 
 export type ApiMode = "live" | "sandbox"
 export type ApiEndpoint = "prepare" | "process" | "task" | "validate" | "verify-signature" | "metadata" | "tracker"
-export type ApiOperation = "send" | "cancel" | "dispense" | "release" | "return" | "withdraw" | "claim"
+export type ApiOperation = "send" | "cancel" | "dispense" | "dispenseamend" |
```
JackRobinsonIW, line 8: does this need to be a single string `dispenseamend` could we do `dispense-amend` instead?

tests/e2e/pact/services/update-prescriptions.ts
```diff
@@ -29,10 +30,13 @@ const x509CertificatePath = process.env.SIGNING_CERT_PATH
 
 const isProd = process.env.APIGEE_ENVIRONMENT === "prod"
 
+let amendBundleIdentifier = null
```
JackRobinsonIW, line 33: don't think we need this to be a top level variable, could sit inside the `updatePrescriptions` function I tihnk

tests/e2e/pact/resources/common.ts
callumtipper-IW: Couldn't do this for the string matching reasons discussed

JackRobinsonIW (approved):

PR #785, opened 2022-02-24
AEA-2081 Test refresh token flow

booshibooshi:


tool/site/server/src/services/session.ts
```diff
@@ -52,3 +54,28 @@ export function removeFromSessionValue(key: string, value: unknown, request: Hap
 export function clearSessionValue(key: string, request: Hapi.Request): void {
   request.yar.clear(key)
 }
+
+export function createSession(tokenResponse: Token, request: Hapi.Request, h: Hapi.ResponseToolkit): void {
+  request.cookieAuth.set({})
+  const accessTokenFetchTime = getUtcEpochSeconds()
+  const refreshTokenTimeout = 3599 / 3
```
JackRobinsonIW, line 61: why are we using this value here as opposed to 1200 or 1199?

tool/site/client/src/components/sessionTimer.tsx
```diff
@@ -14,61 +15,77 @@ export const SessionTimer: React.FC = () => {
...
     }, 1000)
   })
 
-  const timerIntervals = []
+  useEffect(() => {
+    const refreshRequired = getUtcEpochSeconds() > nextRefreshTime
+    if (refreshRequired) {
+      refreshToken().then(result => setNextRefreshTime(result.nextRefreshTime))
+    }
+  })
```
JackRobinsonIW, line 59: split of these two useEffects has me confused, could we have the logic in this lower one run first in the top useEffect? looks like the dependency array is empty for both so the both just run once on render?

tool/site/server/src/services/session.ts
booshibooshi: 3599 is the default for refresh token expiry, so in future we can use the full expiry, but will need some investigation to how to consume the refresh api on the oauth client library, something is mising from data being passed to it which only allows 1 refresh atm

tool/site/client/src/components/sessionTimer.tsx
booshibooshi: consolidated

tool/site/server/src/routes/auth/refresh.ts
```diff
@@ -0,0 +1,23 @@
...
+import {getUtcEpochSeconds} from "../util"
+
+export default {
+  method: "POST",
+  path: "/auth/refresh",
+  handler: async (request: Hapi.Request, h: Hapi.ResponseToolkit): Promise<Hapi.ResponseObject> => {
+    const oauthData = getSessionValue("oauth_data", request)
+    const token = await refreshToken(oauthData)
+    if (token.expired()) {
+      return h.response({redirectUri: `${CONFIG.baseUrl}logout`}).code(400)
```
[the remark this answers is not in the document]
booshibooshi: good shout, updated

JackRobinsonIW (approved):

PR #789, opened 2022-02-28
remove spurious code from secondary care examples

dan-jones-iw:


booshibooshi:
Update removing spurious code has changed author and responsible party correctly, believe there's just the cancel example which needs updating now?

ChrisLopezUK (approved):

PR #803, opened 2022-03-07
Aea 1946 claim amend pact

dan-jones-iw:


tests/e2e/pact/services/update-prescriptions.ts
```diff
@@ -204,6 +194,24 @@ export async function updatePrescriptions(
   })
 }
 
+function isRepeatDispenseNotification(messageHeader: fhir.MessageHeader) {
+  if (messageHeader.extension) {
+    const replacementOfExtension = getReplacementOfExtension(messageHeader.extension)
+    if (replacementOfExtension) {
+      return !!replacementOfExtension.valueIdentifier.value
+    }
+  }
```
JackRobinsonIW, line 203: dont think it really matters but this function return is `void` if there isn't an extension, it should probably return `false` instead

tests/e2e/pact/services/update-prescriptions.ts
```diff
@@ -204,6 +194,24 @@ export async function updatePrescriptions(
...
+}
+
+function isRepeatClaim(claim: fhir.Claim) {
+  if (claim.extension) {
+    const replacementOfExtension = getReplacementOfExtension(claim.extension)
+    if (replacementOfExtension) {
+      return !!replacementOfExtension.valueIdentifier.value
+    }
+  }
+}
```
JackRobinsonIW, line 213: same as previous comment

JackRobinsonIW (approved):

callumtipper-IW (approved):

PR #805, opened 2022-03-09
AEA-1571 Add sandbox EPSAT environment with singular dose to text placeholder page

booshibooshi:


tool/site/server/src/services/communication/signing-client.ts
```diff
@@ -18,6 +18,7 @@ export function getSigningClient(request: Hapi.Request, accessToken: string): Si
   return (isDev(CONFIG.environment) && getSessionValue("use_signing_mock", request))
     || (isQa(CONFIG.environment) && getSessionValue("use_signing_mock", request))
     || isLocal(CONFIG.environment)
+    || isSandbox(CONFIG.environment)
```
booshibooshi, line 21: defunct change, to remove

tool/site/server/src/app.ts
```diff
@@ -45,25 +47,27 @@ function createServer() {
...
-      isSecure: true
-    },
-    redirectTo: (request: Hapi.Request) => {
-      if (isDev(CONFIG.environment)) {
-        setSessionValue(
-          "use_signing_mock",
-          request.query["use_signing_mock"],
-          request
-        )
+  if (!isSandbox(CONFIG.environment)) {
```
JackRobinsonIW, line 50: with the above change (lines 18-20) this function is only called when not in sandbox mode, so not sure we need this check here?

tool/site/server/src/app.ts
```diff
@@ -132,23 +136,30 @@ function addApiRoutes(server: Hapi.Server) {
...
+    server.route(addView("prescribe/cancel"))
+    server.route(addView("dispense/release"))
+    server.route(addView("dispense/verify"))
+    server.route(addView("dispense/return"))
+    server.route(addView("dispense/dispense"))
+    server.route(addView("dispense/withdraw"))
+    server.route(addView("dispense/claim"))
+  }
+
+  if (isDev(CONFIG.environment) || isLocal(CONFIG.environment) || isQa(CONFIG.environment)) {
```
JackRobinsonIW, line 160: I like this change

JackRobinsonIW (dismissed):

tool/site/server/src/app.ts
booshibooshi: Good spot, forgot I'd already made that change some time ago. Will remove duplication

PR #834, opened 2022-03-31
implement diagnostic info for release rejections that require it

dan-jones-iw:


coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -542,3 +533,51 @@ export class ReleaseResponseHandler extends SpineResponseHandler<hl7V3.Prescript
...
+        issue.diagnostics = ReleaseRejectionHandler.getDiagnosticInfo(sendMessagePayload)
+      }
+      return issue
+    })
+    return spineResponse
+  }
+
+  private static issueNeedsDiagnosticInfo(issue: fhir.OperationOutcomeIssue) {
+    const issueDetails = issue.details.coding
+    return issueDetails.some(issue => issue.code === "PRESCRIPTION_WITH_ANOTHER_DISPENSER")
```
JackRobinsonIW, line 563: feel like we should have a const with the value `"PRESCRIPTION_WITH_ANOTHER_DISPENSER"` we could use here rather than the string?

coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -542,3 +533,51 @@ export class ReleaseResponseHandler extends SpineResponseHandler<hl7V3.Prescript
...
+    sendMessagePayload: hl7V3.SendMessagePayload<hl7V3.PrescriptionReleaseRejectRoot>,
+    logger: pino.Logger
+  ): TranslatedSpineResponse {
+    const spineResponse = super.handleErrorResponse(sendMessagePayload, logger)
+    const operationOutcome = spineResponse.fhirResponse as fhir.OperationOutcome
+    operationOutcome.issue.forEach((issue) => {
+      if (ReleaseRejectionHandler.issueNeedsDiagnosticInfo(issue)) {
+        issue.diagnostics = ReleaseRejectionHandler.getDiagnosticInfo(sendMessagePayload)
+      }
+      return issue
```
JackRobinsonIW, line 556: not sure we need this return? its a forEach looping over issues and adding `diagnostics` if needed

if we were using a `.map` to build new issues we'd need the return but as is I think we're good without?

coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -542,3 +533,51 @@ export class ReleaseResponseHandler extends SpineResponseHandler<hl7V3.Prescript
...
+
+    const performerAgentPerson = rejectionReason.performer.AgentPerson
+    const v3Telecom = performerAgentPerson.telecom
+    const firstFhirTelecom = convertTelecom(v3Telecom)[0]
+
+    const v3Org = performerAgentPerson.representedOrganization
+    const odsCode = v3Org.id._attributes.extension
+    const orgName = v3Org.name._text
+
+    const name = {odsCode, name: orgName, tel: firstFhirTelecom.value}
```
JackRobinsonIW, line 580: think we could use a better const name than `name` ;)

JackRobinsonIW (approved):

PR #836, opened 2022-04-05
Included NHSD-Session-URID in Access-Control-Allow-Headers

ChrisLopezUK:
I think I got this, though would appreciate a second pair of eyes.

### DIFF
--- a/proxies/sandbox/apiproxy/policies/AssignMessage.AddCors.xml
+++ b/proxies/sandbox/apiproxy/policies/AssignMessage.AddCors.xml
@@ -4,7 +4,7 @@
     <Set>
         <Headers>
             <Header name="Access-Control-Allow-Origin">{request.header.origin}</Header>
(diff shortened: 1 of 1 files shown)


proxies/sandbox/apiproxy/policies/AssignMessage.AddCors.xml
```diff
@@ -4,7 +4,7 @@
     <Set>
         <Headers>
             <Header name="Access-Control-Allow-Origin">{request.header.origin}</Header>
-            <Header name="Access-Control-Allow-Headers">origin, x-requested-with, accept, content-type, authorization, x-correlation-id, x-request-id</Header>
+            <Header name="Access-Control-Allow-Headers">origin, x-requested-with, accept, content-type, authorization, x-correlation-id, x-request-id, NHSD-Session-URID</Header>
```
dan-jones-iw, line 7: should this be lowercase like the other header values in here?

ChrisLopezUK: Think you are right. Was just looking at the screenshot on try this API.

image

Seems like all mixed case headers have just been made lower. Fixing.

PR #838, opened 2022-04-06
AEA-2143 Add loading spinner to EPSAT

booshibooshi:


tool/site/client/src/components/common/loading.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <StyledLabel isPageHeading>Loading</StyledLabel>
+      <div className="nhsuk-loader">
+        <span className="nhsuk-loader__spinner"></span>
+        <span className="nhsuk-loader__text">{message}</span>
+      </div>
+    </div>
+  )
+}
+
+export const Spinner: React.FC = () => {
```
JackRobinsonIW, line 28: not a huge thing but generally prefer component files to only have a single exported component, could we move this to its own `spinner.tsx` file?

JackRobinsonIW: might have changed my mind actually given how simple the spinner is

tool/site/client/src/pages/loadPage.tsx
```diff
@@ -40,34 +46,45 @@ const LoadPage: React.FC = () => {
...
+
+      async function uploadBundlesInBatchesOfTen(bundles: Bundle[]) {
+        const loadResponses: Array<LoadResponse> = []
+        const chunkSize = 10
+        for (let i = 0; i < bundles.length; i += chunkSize) {
+          const chunk = bundles.slice(i, i + chunkSize)
+          const response = await axiosInstance.post<LoadResponse>(`${baseUrl}prescribe/edit`, chunk)
+          loadResponses.push(getResponseDataIfValid(response, isLoadResponse))
+        }
+        return loadResponses
```
JackRobinsonIW, line 81: super minor but could we use consistent naming here for batch/chunk e.g. its called `uploadBundlesInBatchesOfTen` so line 75 is `const batchSize = 10` and line 77 is `const batch = bundles.slice(i, i + batchSize)` or vice versa

tool/site/client/src/pages/loadPage.tsx
```diff
@@ -40,34 +46,45 @@ const LoadPage: React.FC = () => {
...
       }
     })()
   }, [baseUrl, loadFormValues, prescriptionsInTestPack, prescriptionFilesUploaded, setLoadPageErrors])
```
JackRobinsonIW, line 85: Given the complexity of this page now I think it'd be good to add some unit tests, looks like we don't currently have a `loadPage.spec.ts` file

booshibooshi: I think the complexity of this page is why there weren't unit tests, there is good test-coverage of this page in selenium tests. This is something we might want to cover when we do a review into test-coverage generally, or maybe a new sub-task as part of this work, but separate to this PR

JackRobinsonIW (approved):

PR #839, opened 2022-04-06
update eps to look in claim.contained for legalAuthenticator info

dan-jones-iw:


coordinator/src/services/translation/request/dispense/dispense-claim.ts
```diff
@@ -132,15 +141,25 @@ async function createDispenseClaimPertinentInformation1(
...
 
 async function createLegalAuthenticator(
-  payeeOdsCode: string,
+  practitionerRole: fhir.PractitionerRole,
   timestamp: hl7V3.Timestamp,
   headers: Hapi.Util.Dictionary<string>,
   logger: pino.Logger
 ) {
+  //TODO remove duplication from dispense-notification as part of /Task work
+  const practitionerRoleOrg = practitionerRole.organization as fhir.IdentifierReference<fhir.Organization>
```
JackRobinsonIW, line 150: why do we need this `as` here? I would have thought the type should come through from line 144

callumtipper-IW (approved):

PR #881, opened 2022-05-05
AEA-2247 Add nominated pharmacy and organisation details to test-pack

booshibooshi:


tool/e2e-tests/helpers.ts
```diff
@@ -286,8 +286,10 @@ async function getCreatedPrescriptionId(driver: ThenableWebDriver): Promise<stri
   return prescriptionId
 }
 
+/* eslint-disable @typescript-eslint/no-unused-vars */
 export async function finaliseWebAction(driver: ThenableWebDriver, log: string): Promise<void> {
-  console.log([log, await driver.takeScreenshot()].join("\n"))
+  //console.log([log, await driver.takeScreenshot()].join("\n"))
```
JackRobinsonIW, line 291: If we want to just get rid of the screenshots all together I reckon that's ok, not sure they are adding much value

tool/e2e-tests/prescribe/prescriptionPagination.ts
```diff
@@ -1,63 +1,37 @@
...
-  await driver.wait(until.urlContains(`prescribe/edit?prescription_id=${encodeURIComponent(previousPrescriptionId)}`))
-  await driver.wait(until.elementLocated(sendPageTitle))
+async function loadTheNextPrescription(driver: ThenableWebDriver, nextPageElement: WebElement) {
+  await nextPageElement.click()
 }
 
-function checkPreviousPrescriptionIsTheStartingPrescription(previousPrescriptionId: string, startingPrescriptionId: string) {
-  expect(previousPrescriptionId).toEqual(startingPrescriptionId)
+async function checkPageIsShowingNextPrescription(driver: ThenableWebDriver, nextPageElement: WebElement) {
+  expect((await nextPageElement.getAttribute("class")).split(" ").indexOf("selected") >= 0).toBeTruthy()
```
JackRobinsonIW, line 36: If I'm honest this expect and the one on line 28 have me a little confused about what they are actually checking

booshibooshi: It's checking against a pagination list to ensure that elements are selected as they should be, so this test ensures that we can navigate from page 1 to page 2 successfully

tool/e2e-tests/helpers.ts
booshibooshi: Leaving in for now, as there's opportunity to help with assurance and I don't want us to lose visibility of this yet

tool/site/client/src/components/pagination.tsx
```diff
@@ -0,0 +1,87 @@
...
+          return <li className="pagination-item dots">&#8230;</li>
+        }
+
+        // Render our Page Pills
+        return (
+          <li
+            className={classnames("pagination-item", {
+              selected: pageNumber === currentPage
+            })}
+            onClick={() => onPageChange(pageNumber as number)}
```
JackRobinsonIW, line 68: I think we can get rid of this `as` if we change the type in `usePagination`, I'll leave a comment there

tool/site/client/src/components/prescription-summary/prescriptionLevelDetails.tsx
```diff
@@ -34,6 +34,10 @@ export function createPrescriptionLevelDetails(
   const detailsProps: PrescriptionLevelDetailsProps = {
     prescriptionId,
     courseOfTherapyType: courseOfTherapyTypeCoding.display,
+    prescriptionTypeCode:
+      medicationRequest.extension.find(
+        e => e.url === "https://fhir.nhs.uk/StructureDefinition/Extension-DM-PrescriptionType"
+      ).valueCoding.code,
```
JackRobinsonIW, line 40: any chance this `find` could return `undefined`?

JackRobinsonIW: it shouldn't right, I think that extension is required

tool/site/client/src/components/usePagination.tsx
```diff
@@ -0,0 +1,77 @@
...
+  siblingCount: number
+  currentPage: number
+}
+
+export const usePagination = ({
+  totalCount,
+  pageSize,
+  siblingCount,
+  currentPage
+}: UsePaginiationProps): Array<string | number> => {
```
JackRobinsonIW, line 22: I think we can use `Array<typeof DOTS | number>` then get rid of the casting we do in `pagination.tsx`

tool/site/client/src/pages/signPage.tsx
```diff
@@ -61,76 +49,23 @@ const SignPage: React.FC<SignPageProps> = ({
...
-                  </Table.Row>
-                )}
-              </Table.Body>
-            </Table>
-          </>
-        )
-      }}
-    </LongRunningTask>
-  }
-
```
JackRobinsonIW, line 123: love that this is gone

tool/site/client/src/pages/signPage.tsx
```diff
@@ -61,76 +49,23 @@ const SignPage: React.FC<SignPageProps> = ({
...
-    <LongRunningTask<Bundle> task={retrievePrescriptionDetailTask} loadingMessage="Retrieving prescription details.">
-      {bundle => {
+    <LongRunningTask<Array<Bundle>> task={retrievePrescriptionsTask} loadingMessage="Retrieving prescription details.">
+      {bundles => {
         if (sendPageFormValues.editedPrescriptions.length === 0) {
-          const summaryViewProps = createSummaryPrescriptionViewProps(bundle, editMode, setEditMode)
+          const summaryViewProps = createSummaryPrescriptionViewProps(
+            bundles[currentPage - 1],
+            currentPage,
+            parseInt(Object.keys(bundles).pop()) + 1,
```
JackRobinsonIW, line 59: Is `Object.keys` here an array of numbers as strings e.g. `["1", "2", ... ]`

tool/site/client/src/services/test-packs/index.ts
```diff
@@ -44,31 +52,33 @@ export const createPrescriptionsFromExcelFile = (
...
+  places: Array<fhir.BundleEntry>,
   rows: Array<PrescriptionRow>,
   setLoadPageErrors: Dispatch<SetStateAction<any>>
 ): any[] {
   const prescriptions = []
-  const groupedPrescriptionRows = groupBy(rows, (row: PrescriptionRow) => row.testId)
-  groupedPrescriptionRows.forEach(prescriptionRows => {
+  const prescriptionRows = groupBy(rows, (row: PrescriptionRow) => row.testId)
+
+  prescriptionRows.forEach(prescriptionRows => {
```
JackRobinsonIW, line 65: could we rename either the first or second `prescriptionRows` in `prescriptionRows.forEach(prescriptionRows => {`? It's a bit confusing which is being referred to

tool/site/client/src/services/test-packs/index.ts
```diff
@@ -80,31 +90,37 @@ function createPrescriptions(
...
   patient: fhir.BundleEntry,
   prescriber: fhir.BundleEntry,
+  places: Array<fhir.BundleEntry>,
   prescriptionRows: PrescriptionRow[],
   nominatedPharmacy: string,
   prescriptions: any[]
 ) {
+  const prescriptionRow = prescriptionRows[0]
   const repeatsAllowed = prescriptionRow.repeatsAllowed
   for (let repeatsIssued = 0; repeatsIssued <= repeatsAllowed; repeatsIssued++) {
```
JackRobinsonIW, line 118: just checking the logic here, if `repeatsAllowed` is null because the spread sheet is empty this `for` block will not error it just won't run?

JackRobinsonIW: realized this is existing code so fairly sure it must be working

tool/site/client/src/services/test-packs/placeResources.ts
```diff
@@ -138,11 +148,11 @@ export function createPlaceResources(prescriptionType: PrescriptionType, fhirPre
...
-          line: ["MUSGROVE PARK HOSPITAL"],
-          city: "TAUNTON",
-          postalCode: "TA1 5DA"
+          line: parentOrganisation.address,
+          city: parentOrganisation.city,
+          postalCode: parentOrganisation.postcode
         }
       } as fhir.Location
-    })
+    }]
```
JackRobinsonIW, line 156: we have an `if` and an `else if` but do we need to catch the case in which neither conditional is true?

tool/site/client/static/styles.css
```diff
@@ -96,4 +96,65 @@ border-top-color:#fff
...
+	display: flex;
+	list-style-type: none;
+	justify-content: center;
+}
+.pagination-container .pagination-item {
+	padding: 0 12px;
+	height: 32px;
+	text-align: center;
+	margin: auto 4px;
+	color: rgba(0, 0, 0, 0.87);
```
JackRobinsonIW, line 111: not super important but we could probably use css variables here for color values

tool/site/client/tests/components/prescription-summary/props.tsx
```diff
@@ -54,6 +55,9 @@ export const summaryPrescription: PrescriptionSummaryViewProps = {
   patient: summaryPatient,
   practitionerRole: summaryPractitionerRole,
   prescriptionLevelDetails: prescriptionLevelDetailProps,
+  currentPage: 1,
+  pageCount: 1,
+  onPageChange: page => console.log(page),
```
JackRobinsonIW, line 60: just thinking about keeping the console clear could we not just use a void function here? `page => void(page)`

tool/e2e-tests/helpers.ts
JackRobinsonIW: +1

tool/e2e-tests/prescribe/prescriptionPagination.ts
JackRobinsonIW: Great thank you

tool/site/client/src/components/prescription-summary/prescriptionLevelDetails.tsx
booshibooshi: It's required, but should probably handle the possibility of it missing as users can put in custom prescriptions. Have added defensive coding here

tool/site/client/src/components/usePagination.tsx
booshibooshi: Yep that's much better. Thanks

tool/site/client/src/services/test-packs/placeResources.ts
booshibooshi: these are the only 2 supported values atm. can see where this is set in `test-packs\index.ts function: getPrescriptionType`

tool/site/client/static/styles.css
booshibooshi: will update in future pr

JackRobinsonIW (approved):

PR #882, opened 2022-05-06
Aea separate auth tweaks

dan-jones-iw:


tool/site/server/src/routes/auth/callback.ts
```diff
@@ -131,3 +126,8 @@ export default {
 function getQueryString(query: Hapi.RequestQuery) {
   return Object.keys(query).map(key => `${key}=${query[key]}`).join("&")
 }
+
+function separateAuthLogin(request: Hapi.Request) {
```
JackRobinsonIW, line 130: I'd kinda like this to be called `isSeparateAuthLogin` if we get another failure but wouldn't hold up a merge over it

JackRobinsonIW (approved):

JackRobinsonIW (approved):

PR #886, opened 2022-05-11
Aea 2090 use message telecom if lookup fails

JackRobinsonIW:


coordinator/tests/services/translation/request/agent-unattended.spec.ts
```diff
@@ -96,3 +106,41 @@ const mockOrganizationResponse: fhir.Organization = {
     }
   ]
 }
+
+const mockOrganizationResponseWithoutTelecom: fhir.Organization = {
```
dan-jones-iw, line 110: remove duplication from above

JackRobinsonIW: Done

PR #895, opened 2022-05-17
Aea 2255 omit line item status from digest generation

JackRobinsonIW:


dan-jones-iw (approved):

coordinator/src/services/translation/request/signature.ts
```diff
@@ -33,6 +34,15 @@ function getLineItemWithoutRepeatNumberLow(lineItem: hl7V3.LineItem) {
   }
 }
 
+function getLineItemWithoutStatus(lineItem: hl7V3.LineItem) {
```
dan-jones-iw, line 37: wanna make the above function look similar, think i prefer this style

dan-jones-iw (approved):

PR #902, opened 2022-05-23
removed client-oauth2 from login route

dan-jones-iw:


tool/site/server/src/routes/auth/login.ts
```diff
@@ -20,6 +19,20 @@ interface UnattendedTokenResponse {
...
 }
 
+function getRedirectUri(authorizeUrl: string, clientId: string, callbackUri: string, scopes?: Array<string>) {
+  const queryParams: Record<string, string> = {
+    "client_id": clientId,
+    "redirect_uri": callbackUri,
+    "response_type": "code",
+    "state": uuid.v4()
+  }
+  if (scopes) {
```
JackRobinsonIW, line 29: I generally don't like having optional function parameters if I can avoid it, we could make `scopes` required and then change this to be `if (scopes.length > 0)`, I understand if others disagree though

tool/site/server/src/routes/auth/login.ts
```diff
@@ -85,20 +98,27 @@ export default {
...
     }
 
     // Attended (User)
-    const oauthClient = createOAuthCodeFlowClient()
-    const redirectUri = oauthClient.getUri({state: createOAuthState()})
+    const authorizationUri = `${CONFIG.privateApigeeUrl}/oauth2/authorize`
+    const redirectUri = getRedirectUri(authorizationUri, CONFIG.clientId, callbackUri)
+
+    // const oauthClient = createOAuthCodeFlowClient()
+    // const redirectUri = oauthClient.getUri({state: createOAuthState()})
```
JackRobinsonIW, line 119: can we delete these comments?

JackRobinsonIW (approved):

PR #908, opened 2022-06-07
Aea 2323 release changes

TomSmithCoding:


coordinator/src/services/translation/common/getResourcesOfType.ts
```diff
@@ -72,7 +72,7 @@ export function getLocations(bundle: fhir.Bundle): Array<fhir.Location> {
   return getResourcesOfType<fhir.Location>(bundle, "Location")
 }
 
-function getContainedResource<P extends fhir.Resource, C extends fhir.Resource>(
+function followContainedReference<P extends fhir.Resource, C extends fhir.Resource>(
```
JackRobinsonIW, line 75: minor but I think in other places we have used `resolve{something}Reference` rather than follow

coordinator/src/services/translation/common/index.ts
```diff
@@ -1,7 +1,7 @@
-import {fhir, processingErrors as errors} from "@models"
+import {fhir, processingErrors, processingErrors as errors} from "@models"
```
JackRobinsonIW, line 1: duped import processingErrors?

coordinator/src/services/translation/request/agent-unattended.ts
```diff
@@ -11,49 +11,64 @@ import {
...
-): Promise<hl7V3.Author> {
-  const agentPerson = await createAgentPersonFromPractitionerRole(practitionerRole, logger)
-  const author = new hl7V3.Author()
-  author.AgentPerson = agentPerson
-  return author
+export function createAgentPersonPersonUsingPractitionerRole(
+  practitionerRole: fhir.PractitionerRole,
+): hl7V3.AgentPersonPerson {
+
+  const practitioner = practitionerRole.practitioner as fhir.IdentifierReference<fhir.Practitioner>
```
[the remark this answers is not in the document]
TomSmithCoding: I believe we need to specify the `as` here because the `practitionerRole.practitioner` can either be a type of `Reference `or an `IdentifierReference`, so when attempting to resolve line 62 it was throwing an error due to the possibility of it possibly being a Reference type, if my memory serves right.

examples/secondary-care/community/repeat-dispensing/nominated-pharmacy/clinical-practitioner/single-medication-request/1-Convert-Response-Release-200_OK.xml
```diff
@@ -40,27 +40,28 @@
...
           <AgentPerson classCode="AGNT">
-            <id extension="555254240100" root="1.2.826.0.1285.0.2.0.67"/>
-            <code code="S8000:G8000:R8003" codeSystem="1.2.826.0.1285.0.2.1.104"/>
+            <id extension="555086415105" root="1.2.826.0.1285.0.2.0.67"/>
+            <code code="R8000" codeSystem="1.2.826.0.1285.0.2.1.104"/>
             <telecom use="WP" value="tel:02380798431"/>
             <agentPerson classCode="PSN" determinerCode="INSTANCE">
-              <id extension="555254239107" root="1.2.826.0.1285.0.2.0.65"/>
-              <name>USERQ RANDOM Mr</name>
+              <id extension="1231234" root="1.2.826.0.1285.0.2.1.54"/>
```
dan-jones-iw, line 51: this should be the SDS one (`0.65`)

coordinator/src/services/translation/request/agent-unattended.ts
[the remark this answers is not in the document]
JackRobinsonIW: just wondering if we could use a type guard here then rather than `as` e.g.

```
const practitioner = practitionerRole.practitioner
if (!isReference(practitoner)) {
 throw new Error( wrong type)
}
```

JackRobinsonIW (approved):

PR #913, opened 2022-06-23
AEA-2326 Task message changes

JackRobinsonIW:


coordinator/src/services/translation/request/return/return.ts
```diff
@@ -1,57 +1,53 @@
...
   }
 
+  const practitionerRole: fhir.PractitionerRole = getContainedPractitionerRole(
+    task,
+    task.requester.reference
+  )
+
+  if (!isReference(practitionerRole.organization)) {
+    throw new errors.InvalidValueError(
+      "For return messages, practitionerRole.organization must be a reference to a contained Organization resource."
```
dan-jones-iw, line 30: i know this is copied from line 19 but this code will run for both `return` and `withdraw` messages

PR #917, opened 2022-06-30
Added JSON view component to EPSAT

JackRobinsonIW:


tool/site/client/package-lock.json
```diff
@@ -13,6 +13,8 @@
         "nhsuk-react-components-extensions": "^1.5.4",
         "react-cookie": "^4.1.1",
         "react-diff-viewer": "^3.1.1",
+        "react-json-pretty": "^2.2.0",
```
mukhtar-abdul-ghani, line 16: this needs to be deleted

JackRobinsonIW: done

tool/site/client/src/components/messageExpanders.tsx
```diff
@@ -21,7 +22,8 @@ const MessageExpanders: React.FC<MessageExpandersProps> = ({
   <Details.ExpanderGroup>
     <MessageExpander
       name="Request (FHIR)"
-      message={JSON.stringify(fhirRequest, null, 2)}
+      message={JSON.stringify(fhirResponse, null, 2)}
```
dan-jones-iw, line 25: should this be `fhirRequest`

tool/site/client/tests/pages/validatePage.spec.tsx
```diff
@@ -31,8 +31,8 @@ test("Displays validate result", async () => {
     status: 200,
     response: {
       success: true,
-      request: "JSON Request",
-      response: "JSON Response"
+      request: {req: "JSON Request"},
```
dan-jones-iw, line 34: why are these changes being made (lines 34-35 & 45-46)?

tool/site/client/src/components/messageExpanders.tsx
```diff
@@ -45,6 +48,7 @@ const MessageExpanders: React.FC<MessageExpandersProps> = ({
 interface MessageExpanderProps {
   name: string
   message: string
```
dan-jones-iw, line 50: kinda of annoying we are passing in basically the same thing twice, just one typed. can we make do without the prop `message`?

tool/site/client/src/components/messageExpanders.tsx
JackRobinsonIW: Good spot

tool/site/client/tests/pages/validatePage.spec.tsx
JackRobinsonIW: With the previous implementation we basically called `JSON.stringify` on the `request` and `response` fields. This was a little weird but didn't error as `JSON.stringify` can accept a string as a param

The new view component however needs to receive an object so we need this mock to return an `object` rather than simply a `string`

tool/site/client/src/components/messageExpanders.tsx
JackRobinsonIW: the `message` prop has to exist as we need to pass the `xml` data as a string, I could get rid of the `jsonMessage` potentially and have something like:

```
let displayMessage: FhirResponse | string

try {
  displayMessage = JSON.parse(message)
} catch {
  displayMessage = message
}

// render message
{ typeOf(message) === string ? <xml string> : <json view component> }
```

JackRobinsonIW: changed this to be two distinct components

tool/site/client/src/components/messageExpanders.tsx
```diff
@@ -19,40 +20,94 @@ const MessageExpanders: React.FC<MessageExpandersProps> = ({
...
 
-const StyledButton = styled(Button)`
-  margin-bottom: 0;
-`
+export const JsonMessageExpander: React.FC<JsonMessageExpanderProps> = ({
+  name,
+  message,
+  mimeType
+}) => {
+  const downloadHref = `data:${mimeType};charset=utf-8,${encodeURIComponent(JSON.stringify(message))}`
```
dan-jones-iw, line 68: can we have `JSON.stringify(message, null, 2)` to format nicely

tool/site/client/src/components/common/rawApiResponse.tsx
```diff
@@ -73,10 +73,10 @@ const RawApiResponse: React.FC<RawApiResponseProps> = ({
           </SummaryList.Row>
         </SummaryList>
         {response.headers && (
-          <XmlMessageExpander name="Response Headers" message={formatAsString(response.headers)} mimeType="text/plain"/>
+          <XmlMessageExpander name="Response Headers" message={formatAsString(response.headers)}/>
```
dan-jones-iw, line 76: this used to be `text/plain`, now its `text/xml`, is this a problem?

tool/site/client/src/components/messageExpanders.tsx
```diff
@@ -19,45 +20,90 @@ const MessageExpanders: React.FC<MessageExpandersProps> = ({
...
+        </ButtonList>
+        <ReactJson
+          collapseStringsAfterLength={50}
+          displayDataTypes={false}
+          displayObjectSize={false}
+          name={false}
+          src={message}
+          style={{marginTop: "10px"}}
+          theme={
+            {
```
dan-jones-iw, line 77: would u mind making this a separate const outside of this scope just for readability and stuff

tool/site/client/src/components/common/rawApiResponse.tsx
JackRobinsonIW: Just done some testing, this is continuing to work fine

tool/site/client/src/components/messageExpanders.tsx
JackRobinsonIW: moved

PR #920, opened 2022-07-04
Addtional instructions in EPSAT

JackRobinsonIW:


tool/site/client/src/services/test-packs/xls.ts
```diff
@@ -187,7 +188,7 @@ export function parsePrescriptionRows(rows: Array<XlsRow>, setLoadPageErrors: Di
 
   return rows.map(row => {
     const prescriptionType = row["Prescription Type"]
-    return {
+    const rowData: PrescriptionRow = {
```
dan-jones-iw, line 191: might as well inline

dan-jones-iw: no other changes tho so no biggie

PR #922, opened 2022-07-05
Update test pack organisations

booshibooshi:


coordinator/src/routes/index.ts
```diff
@@ -2,7 +2,7 @@ import convertPrescriptionRoutes from "./debug/convert"
 import validatorRoutes from "./debug/validate"
 import doseToTextRoutes from "./debug/dose-to-text"
 import preparePrescriptionRoutes from "./prescribe/prepare"
-import processPrescriptionRoutes from "./process"
+import processPrescriptionRoutes from "./prescribe/process"
```
dan-jones-iw, line 5: think this should stay at top level since `process-message` serves all of: prescription-order, prescription-order-update and dispense-notification

tool/e2e-tests/test-packs/test-packs.ts
```diff
@@ -4,15 +4,24 @@ import path from "path"
...
 
 export async function loadSupplierTestPack1Examples(driver: ThenableWebDriver): Promise<void> {
+  await loadTestPack(driver, "Supplier 1 Test Pack.xlsx")
+}
+
+export async function loadPrescriptionTypeTestPack1Examples(driver: ThenableWebDriver): Promise<void> {
+  await loadTestPack(driver, "Prescription Types Test Pack.xlsx")
+}
+
+export async function loadPrescriptionTypeTestPack2Examples(driver: ThenableWebDriver): Promise<void> {
```
dan-jones-iw, line 18: is it worth naming these a bit more specifically? rather than TP1/TP2, should we match the name of the file (so for example `loadBadScriptTypesTestPack` or something)

tool/site/client/src/services/test-packs/index.ts
```diff
@@ -52,38 +63,47 @@ export const createPrescriptionsFromExcelFile = (
 }
 
 function createPrescriptions(
-  prescriptionType: PrescriptionType,
-  patients: Array<fhir.BundleEntry>,
-  prescribers: Array<fhir.BundleEntry>,
-  places: Array<fhir.BundleEntry>,
-  rows: Array<PrescriptionRow>,
+  patientRows: Map<string, PatientRow>,
```
dan-jones-iw, line 66: very nice improvements here :+1:

tool/site/client/src/services/test-packs/index.ts
```diff
@@ -208,6 +236,12 @@ function createPrescription(
   return fhirPrescription
 }
 
+function pad(value: string, size: number) {
+  value = value.toString()
+  while (value.length < size) value = "0" + value
```
dan-jones-iw, line 241: just double checking 0s are at the front and not the end for ods codes? should this padding apply when you are talking about a 3-character ods code (parent org) and a 5/6-character ods code (child org)?

booshibooshi: it's from the prescription type scenarios we have to prove out. the short form prescription id needs to contain the org code in the middle of it. in primary care the ods code should be the correct length as that's what the hl7v3 api was built around, but in other care settings ods codes can be (and usually are) shorter, for these cases we need to 0 pad from the start of the odscode so that the short form prescription id is valid

tool/site/client/src/services/test-packs/index.ts
booshibooshi: thanks, I've separated the concerns for data population and building fhir resources as they were previously mixed in various functions before which made it very hard to follow

tool/site/client/src/services/test-packs/index.ts
dan-jones-iw: cool just wanted to make sure there was a clear reason

tool/e2e-tests/test-packs/test-packs.ts
```diff
@@ -3,16 +3,25 @@ import {apiTimeout, defaultWaitTimeout} from "../helpers"
...
+
+export async function loadSupplierTestPack(driver: ThenableWebDriver): Promise<void> {
+  await loadTestPack(driver, "Supplier 1 Test Pack.xlsx")
+}
+
+export async function loadPrescriptionTypeTestPack(driver: ThenableWebDriver): Promise<void> {
+  await loadTestPack(driver, "Prescription Types Test Pack.xlsx")
+}
+
+export async function loadPrescriptionTypesWithInvalidTypesTestPack(driver: ThenableWebDriver): Promise<void> {
```
dan-jones-iw, line 18: these function names are super useful and friendly to read - now tempted to rename the test packs themselves to match these names. approved anyway cause not a big deal

dan-jones-iw (approved):

PR #937, opened 2022-07-14
AEA-2329-Remove userinfo work from proxies

TomSmithCoding:


proxies/live/apiproxy/targets/apim.xml
```diff
@@ -44,18 +44,6 @@
           <Condition>(access-mode Equals "user-restricted")</Condition>
         </Step>
         <!-- Get ODS code from UserInfo -->
-        <Step>
-          <Name>JavaScript.GetUserDetailsFromCache</Name>
-          <Condition>(access-mode Equals "user-restricted")</Condition>
-        </Step>
-        <Step>
-          <Name>AssignMessage.AddUserInfoHeaders</Name>
```
dan-jones-iw, line 52: should we look in the eps coordinator to find where these headers were consumed? probs needs deleting around there somewhere too

coordinator/src/services/validation/parameters-validator.ts
```diff
@@ -25,18 +24,12 @@ export function verifyParameters(
 
   const incorrectValueErrors = []
 
-  const ownerParameter = getOwnerParameter(parameters)
-  if (ownerParameter) {
-    const bodyOrg = getIdentifierValueForSystem(
-      ownerParameter.resource.identifier,
-      "https://fhir.nhs.uk/Id/ods-organization-code",
-      `Organization.identifier`
+  const resourceParameters = parameters.parameter.filter(isResourceParameter)
```
dan-jones-iw, line 27: why have we made this change? cant we stick with `getOwnerParameter`?

coordinator/src/services/validation/parameters-validator.ts
```diff
@@ -25,18 +24,12 @@ export function verifyParameters(
...
+  const ownerParameter = resourceParameters.find(parameter => parameter.name === "owner")
+  if(!ownerParameter){
```
dan-jones-iw, line 29: whitespace :)

coordinator/src/services/validation/parameters-validator.ts
TomSmithCoding: Yesterday you mentioned writing some code here instead of just having the "getOwnerParameter" line by itself, so that it was clear what that line was doing when it was throwing an error for the test, is this not what you meant?

TomSmithCoding: Because the error is thrown inside of "getOwnerParameter"

coordinator/src/services/translation/common/index.ts
```diff
@@ -259,7 +259,7 @@ export function isIdentifierParameter(parameter: fhir.Parameter): parameter is f
   return (parameter as fhir.IdentifierParameter).valueIdentifier !== undefined
 }
 
-function isResourceParameter<R extends fhir.Resource>(
+export function isResourceParameter<R extends fhir.Resource>(
```
dan-jones-iw, line 262: shouldnt need this export anymore

dan-jones-iw (approved):

PR #941, opened 2022-07-25
Aea 2330 log discrepancies between token and message

TomSmithCoding:


coordinator/src/services/validation/parameters-validator.ts
```diff
@@ -41,5 +49,33 @@ export function verifyParameters(
...
 
+  if (practitioner && !isReference(practitioner)) {
+    const bodySDSUserID = getIdentifierValueForSystem(
+      [practitioner.identifier],
+      "https://fhir.nhs.uk/Id/sds-user-id",
+      'parameters.parameter("PractitionerRole").practitioner.identifier'
+    )
+    if (bodySDSUserID !== accessTokenSDSUserID) {
+      console.warn(
+        `PractitionerRole details do not match in
```
dan-jones-iw, line 60: can we be a bit more specific in this log? something like 
`SDS User ID does not match between message body and access token. Body: ${bodySDSUserID}. AccessToken: ${accessTokenSDSUserId}.`

dan-jones-iw (approved):

PR #949, opened 2022-07-29
stop smoke tests from failing silently

dan-jones-iw:


tests/e2e/pact/broker/verify.ts
```diff
@@ -141,12 +141,15 @@ async function clearData() {
     .filter(isNominatedRelease)
 
   for (const nominatedReleaseRequest of nominatedReleaseRequests) {
```
booshibooshi, line 143: Do we need this clearing out of releases anymore if all our release examples have been moved to patient release?

mukhtar-abdul-ghani: We agree with you and will make those changes. We want to improve the smoke tests so they don't return any more false positives. Maybe something we can do to improve the smoke tests in a separate pr?

booshibooshi (dismissed):

tests/e2e/pact/broker/verify.ts
booshibooshi: Sounds good, I've approved these changes. So will leave with you whether you want to merge as is now and make the change on a separate PR, or push new changes here and I'll re-review

booshibooshi (approved):

PR #961, opened 2022-08-10
PrescriptionType work - Release response changes

TomSmithCoding:


coordinator/src/services/translation/response/common.ts
```diff
@@ -196,7 +196,30 @@ export function translateAgentPerson(agentPerson: hl7V3.AgentPerson): Translated
       locations
     }
   } else {
-    if (isSecondaryCare(representedOrganization)) {
+    if (prescriptionType?.startsWith("01", 0)) {
```
StefanoFrazzetto, line 199: Could this check be done in a function that makes it clear what is being checked? E.g. something like `isPrimaryOrSeconaryCare` - it makes it easier, for people who don't have much context, to understand what the code is doing

StefanoFrazzetto (commented):
The code looks good to me - only two comments about minor things. I'll leave the rest of the review to @dan-jones-iw

coordinator/src/services/translation/response/common.ts
dan-jones-iw: `shouldHavePrimaryCareFormat`?

coordinator/tests/services/translation/response/common.spec.ts
```diff
@@ -4,12 +4,14 @@ import {
...
   generateResourceId,
-  getFullUrl
+  getFullUrl,
+  translateAgentPerson
 } from "../../../../src/services/translation/response/common"
 import {fhir, hl7V3} from "@models"
 import {getMedicationRequests} from "../../../../src/services/translation/common/getResourcesOfType"
 import {getExtensionForUrl, resolveReference} from "../../../../src/services/translation/common"
 import {clone} from "../../../resources/test-helpers"
+import {agentPerson} from "../../../resources/test-data/hl7V3"
```
dan-jones-iw, line 14: should we try importing all test data from the `test-data` module?

coordinator/src/services/translation/response/common.ts
TomSmithCoding: Added

coordinator/src/services/translation/response/common.ts
```diff
@@ -228,34 +251,19 @@ export function translateAgentPerson(agentPerson: hl7V3.AgentPerson): Translated
...
-        organization
-      }
-
       return translatedAgentPerson
     }
   }
 }
 
+function shouldHavePrimaryCareFormat(prescriptionType?: string): boolean {
+  if (prescriptionType?.startsWith("01", 0)) {
```
dan-jones-iw, line 260: this can be simplified a lot
`return prescriptionType?.startsWith("01", 0)`

TomSmithCoding: My bad, lazy c&p

dan-jones-iw (approved):

PR #973, opened 2022-09-05
AEA-2704 Bugfix for mandatory sds user ID, no longer mandatory and new test added

TomSmithCoding:


coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -129,12 +129,12 @@ export function verifyCommonBundle(
...
       const practitioner = resolveReference(bundle, practitionerRole.practitioner)
       if (practitioner) {
-        const bodySDSUserID = getIdentifierValueForSystem(
+        const bodySDSUserID = getIdentifierValueOrNullForSystem(
           practitioner.identifier,
           "https://fhir.nhs.uk/Id/sds-user-id",
           'Bundle.entry("Practitioner").identifier'
         )
-        if (bodySDSUserID !== accessTokenSDSUserID) {
+        if (bodySDSUserID && bodySDSUserID !== accessTokenSDSUserID) {
```
StefanoFrazzetto, line 137: Could you please add a variable or a comment about what we're checking here? E.g. `const isConditionWeAreChecking = bodySDSUserID && bodySDSUserID !== accessTokenSDSUserID`

StefanoFrazzetto (dismissed):

PR #977, opened 2022-09-08
Aea 2560 remove organisation type in cancel response

AndyMather24:
createOrganization function now creates Organisation with the type field not populated by default. The createOrganization method is used to populate organisation for the cancel response and release response. Tests updated appropriately.


TomSmithCoding (approved):
Good work Andy!

PR #978, opened 2022-09-09
AEA-2702 End-to-end logging improvements

StefanoFrazzetto:
- Added a new log point for request payload identifier
- Refactored payload creation into its own module

**Request & payload identifiers:**
- X-Request-ID is mapped to the ebXML message's MessageId and ConversationId (as the latter is logged with MWS0032 in Spine, along with the payload ID)
- FHIR payloads' identifier is still mapped to the id root attributes of the HL7 v3 message


TomSmithCoding (approved):

coordinator/src/routes/util.ts
```diff
@@ -149,6 +149,7 @@ export function externalValidator(handler: Hapi.Lifecycle.Method) {
     const showWarnings = getShowValidationWarnings(request.headers) === "true"
     const fhirValidatorResponse = await getFhirValidatorErrors(request, showWarnings)
     if (fhirValidatorResponse) {
+      request.logger.error(`FHIR validator failed to with errors: ${fhirValidatorResponse}`)
```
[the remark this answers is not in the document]
StefanoFrazzetto: I'm not familiar with validator, but it looks like only exceptions are logged: https://github.com/NHSDigital/validation-service-fhir-r4/blob/023f2da89ffe983d51fed699d95e9d756ed219a2/src/main/kotlin/com/example/fhirvalidator/controller/ValidateController.kt#L47

Anyway, it's better having an extra log than no logs

coordinator/src/services/translation/request/payload/factory.ts
```diff
@@ -0,0 +1,203 @@
...
+
+type Payload<T extends PayloadContent> = {
+  content: T,
+  interactionId: hl7V3.Hl7InteractionIdentifier
+}
+
+type FactoryInput = fhir.Bundle | fhir.Task | fhir.Parameters | fhir.Claim
+
+export abstract class PayloadFactory {
+  protected abstract create(
```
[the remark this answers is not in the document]
StefanoFrazzetto: `create` is only used within the `PayloadFactory` class, so I don't think that would be the case?

coordinator/src/services/translation/request/payload/factory.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  static forParameters(): PayloadFactory {
+    return new ParametersTranslationResultFactory()
+  }
+
+  static forClaim(): PayloadFactory {
+    return new ClaimTranslationResultFactory()
+  }
+
+  public makeSendMessagePayload(
```
dan-jones-iw, line 68: in a similar vein to the comment above, `create` is used in a method called `makeSendMessagePayload`, so `make` might be a better word?

coordinator/src/services/translation/request/payload/factory.ts
```diff
@@ -0,0 +1,203 @@
...
+    const logObject = {
+      payloadId: this.getPayloadId(fhirResource)
+    }
+    logger.info(logObject, "Creating Spine payload from FHIR resource")
+  }
+}
+
+class BundleTranslationResultFactory extends PayloadFactory {
+  getPayloadId(bundle: fhir.Bundle): string {
+    return getMessageIdFromBundle(bundle)
```
dan-jones-iw, line 94: is there a conceptual difference between `payloadId` and `messageId`? if not, can we sync up the names?

coordinator/src/services/translation/request/payload/factory.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  private createDispenserWithdraw(task: fhir.Task): hl7V3.EtpWithdrawRoot {
+    const etpWithdraw = convertTaskToEtpWithdraw(task)
+    return new hl7V3.EtpWithdrawRoot(etpWithdraw)
+  }
+}
+
+class ParametersTranslationResultFactory extends PayloadFactory {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getPayloadId(parameters: fhir.Parameters): string {
```
dan-jones-iw, line 171: im not sure typescript will complain if you don't provide all parameters? (could be wrong)

coordinator/tests/services/communication/spine-communication.spec.ts
```diff
@@ -16,6 +16,14 @@ describe("Spine communication", () => {
 
   const logger = pino()
 
+  const mockRequest = {
+    message: "test",
+    interactionId: "test2",
+    messageId: "DEAD-BEEF",
```
dan-jones-iw, line 22: ? lol

coordinator/src/services/translation/request/payload/factory.ts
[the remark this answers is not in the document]
StefanoFrazzetto: What I meant is that the method is not used anywhere else, aside from the unexported classes within the same module, so I didn't think it could cause any confusion. Anyway, since that's not the case, I've renamed it to `createPayload`, so the meaning should be clear now

coordinator/src/services/translation/request/payload/factory.ts
StefanoFrazzetto: Good spot-- `payloadId` is the identifier for the payload, while `messageId`s are generated by Apigee. I'll update those functions

coordinator/src/services/translation/request/payload/factory.ts
StefanoFrazzetto: Added that because of linter warnings

coordinator/tests/services/communication/spine-communication.spec.ts
StefanoFrazzetto: Just a recognisable hex string :)
https://stackoverflow.com/questions/2907262/what-does-dead-beef-mean

dan-jones-iw (approved):

PR #995, opened 2022-10-13
AEA-2756-nominated-pharmacy-type-added Fixing the smoke test failure 

MathewArnoldIW:
… performer value of 4 as well as 0004 from excel sheets


tool/site/client/src/services/test-packs/xls.ts
```diff
@@ -215,6 +215,7 @@ export interface PrescriptionRow {
   issueDurationInDays: string
   dispenserNotes: Array<string>
   nominatedPharmacy?: string
+  nominatedPharmacyType: string
```
AndyMather24, line 218: Hey @MathewArnoldIW,  Is this correct that the nominatedPharmacyType can't be nullable? It looks like the nominatedPharmacy can be null. I'm thinking you cant have a nominatedPharmacyType if you haven't got a nominatedPharmacy? 

give me a shout if you want to discuss any of this?

MathewArnoldIW: Discussed and agreed with Andy

AndyMather24 (approved):

PR #998, opened 2022-10-24
Updated determinerCode from KIND to INSTANCE

TomSmithCoding:
Currently when the FHIR Dispense Notification is being translated to HL7v3, the Supplied Mateterial deteminer code in the HL7v3 message is showing the incorrect value of 'KIND', rather than the expected value of 'INSTANCE'.

Updated the determinerCode and ran the conversions.spec and prepare.spec to update the relevant examples (hence the amount of file changes, DON'T BE INTIMIDATED BY THE AMOUNT OF FILES)

Going to check against PR EPSAT to see if appropriate change has been made without breaking any other HL7 translations.

Can demo for review if needed.


StefanoFrazzetto (approved):
Good work- thanks again for the walk-through!

PR #1010, opened 2022-11-09
AEA-2794 Migrate Pact broker to AWS

StefanoFrazzetto:
Postponed until we have time to deal with it.


MatthewPopat-NHS:
We decided to publish files with the pacts rather than using a pact broker.

PR #1020, opened 2022-11-15
AEA-2749 Verify on release response structure

MatthewPopat-NHS:


AndyMather24 (approved):

coordinator/src/services/translation/response/release/release-response.ts
```diff
@@ -1,32 +1,92 @@
...
+      id: uuid.v4(),
+      meta: {
+        lastUpdated: convertHL7V3DateTimeToIsoDateTimeString(releaseResponse.effectiveTime)
+      },
+      identifier: {
+        system: "https://tools.ietf.org/html/rfc4122",
+        value: releaseResponse.id._attributes.root.toLowerCase()
+      },
+      type: "collection",
+      total: failedPrescriptions.length * 2,
```
StefanoFrazzetto, line 80: The result is the same, but wouldn't it be better to have `passedPrescriptions.length + failedPrescriptions.length`?

coordinator/src/services/translation/response/release/release-response.ts
```diff
@@ -1,32 +1,92 @@
...
+        system: "https://tools.ietf.org/html/rfc4122",
+        value: releaseResponse.id._attributes.root.toLowerCase()
+      },
+      type: "collection",
+      total: passedPrescriptions.length,
+      entry: passedPrescriptions
+        .map(prescriptionResult => createInnerBundle(prescriptionResult.prescription, releaseRequestId))
+        .map(convertResourceToBundleEntry)
+    }
+  }
```
StefanoFrazzetto, line 66: Could we move the creations of passed and failed params creation to separate functions? It would make it easier to understand what `createOuterBundle` does under the hood

StefanoFrazzetto (commented):
Nice job! I left a few comments about minor things that could be done to improve readability/maintainability of the code

coordinator/src/services/translation/response/release/release-response.ts
```diff
@@ -1,32 +1,92 @@
...
+      },
+      expression: ["Provenance.signature.data"]
+    }]
+  }
+}
+
+export function createOuterBundle(releaseResponse: hl7V3.PrescriptionReleaseResponse): fhir.Parameters {
   const releaseRequestId = releaseResponse.inFulfillmentOf.priorDownloadRequestRef.id._attributes.root
   const parentPrescriptions = toArray(releaseResponse.component)
     .filter(component => component.templateId._attributes.extension === SUPPORTED_MESSAGE_TYPE)
```
StefanoFrazzetto, line 41: Just spotted this- I wonder if it would be a good idea to create a separate function to do this check and add some comments with a bit of context, e.g. why do we only care about this message type?

StefanoFrazzetto (approved):

PR #1022, opened 2022-11-17
Aea 2648 verify chain of cert

AndyMather24:
check cert with subCa with unit tests
Updated pipeline to use INT cert when int env
TODO: update pipeline to use prod cert when prod (need confident int works first)


AndyMather24:
not ready

PR #1024, opened 2022-11-22
AEA-2647 verify at time of signature

AndyMather24:
Checking the signature date is in range of the cert start and end date

### DIFF
--- a/coordinator/src/services/verification/signature-verification.ts
+++ b/coordinator/src/services/verification/signature-verification.ts
@@ -6,7 +6,7 @@ import {createParametersDigest} from "../translation/request"
 import crypto from "crypto"
 import {isTruthy} from "../translation/common"
 import {convertHL7V3DateTimeToIsoDateTimeString, isDateInRange} from "../translation/common/dateTime"
-function verifySignature(parentPrescription: hl7V3.ParentPrescription): Array<string> {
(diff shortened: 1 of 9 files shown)


coordinator/src/services/verification/signature-verification.ts
```diff
@@ -5,7 +5,7 @@ import {convertFragmentsToHashableFormat, extractFragments} from "../translation
 import {createParametersDigest} from "../translation/request"
 import crypto from "crypto"
 import {isTruthy} from "../translation/common"
-
+import {convertHL7V3DateTimeToIsoDateTimeString, isDateInRange} from "../translation/common/dateTime"
 function verifySignature(parentPrescription: hl7V3.ParentPrescription): Array<string> {
```
StefanoFrazzetto, line 9: As mentioned earlier, we could rename this to `verifyPrescriptionSignature`

MatthewPopat-NHS (changes requested):

coordinator/src/services/verification/signature-verification.ts
AndyMather24: thanks good suggestion method renamed

MatthewPopat-NHS (dismissed):

MatthewPopat-NHS (approved):

PR #1026, opened 2022-11-23
Aea-2749 Verify on release logging

MatthewPopat-NHS:


coordinator/src/services/translation/response/release/release-response.ts
```diff
@@ -60,24 +61,34 @@ function createPrescriptionsBundleParameter(
...
+  logger: pino.BaseLogger): fhir.Parameters {
   const releaseRequestId = releaseResponse.inFulfillmentOf.priorDownloadRequestRef.id._attributes.root
   const parentPrescriptions = toArray(releaseResponse.component)
     .filter(component => component.templateId._attributes.extension === SUPPORTED_MESSAGE_TYPE)
     .map(component => ({
       prescription: component.ParentPrescription,
       errors: verifySignature(component.ParentPrescription)
     }))
-  const passedPrescriptions = parentPrescriptions.filter(prescriptionResult => prescriptionResult.errors.length === 0)
+  const passedPrescriptionResources = parentPrescriptions
```
AndyMather24, line 74: Hi @MatthewPopat-NHS looks like we iterate over parentPrescriptions twice when we are filtering passed and failed prescriptions. If we used a parentPrescriptions.reduce() we could return an object which has a failedList and a passedList and only iterate over once.

I'm happy to talk about this on a call to explain further

coordinator/tests/services/translation/response/release/release-response.spec.ts
```diff
@@ -29,7 +29,17 @@ import {Organization as IOrgansation} from "../../../../../../models/fhir/practi
 
 describe("outer bundle", () => {
   describe("passed prescriptions", () => {
-    const result = translateReleaseResponse(getExamplePrescriptionReleaseResponse("release_success.xml"))
+    const logger = {
```
AndyMather24, line 32: Hi @MatthewPopat-NHS, to reduce repeating ourselves we could pull the mock logger out of the each describe block. 

We could have a file responsible for mock loggers which can then be used by more tests in the future

const mockErrorLogger =  {
      level: "error",
      info: jest.fn(),
      error: jest.fn(),
      warn: jest.fn(),
      debug: jest.fn(),
      fatal: jest.fn(),
      trace: jest.fn(),
      silent: jest.fn()
    }

If someone needs an mockInfoLogger for a test in future they can added 


const _mockinfoLogger_ =  {
      level: "info",
      info: jest.fn(),
      error: jest.fn(),
      warn: jest.fn(),
      debug: jest.fn(),
      fatal: jest.fn(),
      trace: jest.fn(),
      silent: jest.fn()
    }

Or even a mockLogerFactory method that creates the mock logger with which ever level we need

coordinator/src/services/translation/response/release/release-response.ts
```diff
@@ -60,24 +61,34 @@ function createPrescriptionsBundleParameter(
...
+    .filter(prescriptionResult => prescriptionResult.errors.length === 0)
```
AndyMather24, line 75: Hi @MatthewPopat-NHS looks like we iterate over parentPrescriptions twice when we are filtering passed and failed prescriptions. If we used a parentPrescriptions.reduce() we could return an object which has a failedList and a passedList and only iterate over once.

I'm happy to talk about this on a call to explain further

AndyMather24 (approved):
Great work!

PR #1027, opened 2022-11-28
AEA-2777 Create prescription for Spine tracker tests in CI

StefanoFrazzetto:
The Spine tracker Pact test is currently failing in Ref because the prescription does not exist.
This should fix the issue by creating a prescription at run time, which will be then retrieved by the test.


anthony-nhs:
not needed

PR #1028, opened 2022-11-28
Aea 2648 verify chain of cert

AndyMather24:
Read level 1D cert from AWS secret manager and check cert of prescription was issued. 
Including unit tests


packages/coordinator/src/services/verification/signature-verification.ts
```diff
@@ -32,6 +32,11 @@ function verifyPrescriptionSignature(parentPrescription: hl7V3.ParentPrescriptio
   return errors
 }
 
+function verifyChain(x509Certificate: crypto.X509Certificate): boolean {
+  const x509CertificateRoot = new crypto.X509Certificate(process.env.SUBCACC_CERT)
+  return x509Certificate.checkIssued(x509CertificateRoot)
+}
+
```
StefanoFrazzetto, line 39: As already mentioned: should we check Sub CA 1D against the Root CA, and the certificate within the prescription against the Sub CA 1D certificate? We also need to make sure we document this decision within the ticket, for future reference.

StefanoFrazzetto (commented):
Great work so far; your code looks nice!

Let's clarify those few comments, including what we've discussed this morning about the new cert, and then we can have another look.

packages/coordinator/tests/services/verification/signature-verification.spec.ts
```diff
@@ -1,165 +1,200 @@
...
+    })
+    test("should not return error message when cert has not expired", () => {
+      const result = verifyCertificate(parentPrescription)
+      const certificateHasExpired = result.includes(certExpiredErrorMessage)
+      expect(certificateHasExpired).toBeFalsy()
+    })
+  })
+
+  describe("VerifyChain", () => {
+    const createX509Cert = (certPath: string): X509Certificate => {
```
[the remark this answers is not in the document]
AndyMather24: Yeah good shout getX509Cert is better name

packages/coordinator/src/services/verification/signature-verification.ts
AndyMather24: Hi @StefanoFrazzetto, I have renamed the dummy certs in the repo as the names were causing confusion. There was one called root which wasn't a root and the trustedX509 is a mock cert i created which is only trusted by the mock subCa created it is not trusted by the INT subCa. However the tests covering the release response use prescriptions signed by an int smart card which why I have INT subCa. 

There is a comment from Tim on the ticket regarding just checking against the subCa. He is also raising a KAD to confirm the decision. 

We can catch up on this if required when  you are back.

packages/coordinator/package.json
```diff
@@ -26,6 +26,7 @@
     "@hapi/boom": "^10.0.0",
     "@hapi/hapi": "^20.2.2",
     "axios": "^0.21.1",
+    "build": "^0.1.4",
```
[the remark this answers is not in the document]
AndyMather24: I'm not sure why this is here and it is not required.  I have removed it.

packages/coordinator/src/services/verification/signature-verification.ts
```diff
@@ -32,6 +32,28 @@ function verifyPrescriptionSignature(parentPrescription: hl7V3.ParentPrescriptio
   return errors
 }
 
+function verifyChain(x509Certificate: crypto.X509Certificate): boolean {
+  const subCACerts = getSubCaCerts()
+  return certIsTrusted(x509Certificate, subCACerts)
+}
+
+const getSubCaCerts = (): Array<string> => {
```
MathewArnoldIW, line 40: could be replaced with a simple line as split() will always return an array.

packages/coordinator/tests/resources/test-helpers.ts
```diff
@@ -59,8 +61,16 @@ declare global {
 }
 
 if (!Array.prototype.remove) {
-  Array.prototype.remove = function<T>(this: Array<T>, elem: T): void {
+  Array.prototype.remove = function <T>(this: Array<T>, elem: T): void {
```
MathewArnoldIW, line 64: This space may have been accidentally added

packages/coordinator/src/services/verification/signature-verification.ts
AndyMather24: great point. done

packages/coordinator/tests/resources/test-helpers.ts
AndyMather24: removed

MathewArnoldIW (dismissed):

PR #1034, opened 2022-11-30
AEA-2826 Swapped rate limiting to use built-in APIM flow and switched to per minute spike arrest

MatthewPopat-NHS:
(per second spike arrests are amortized over milliseconds whereas per minute spike arrests are amortized over seconds)


MatthewPopat-NHS:
Closing this one as we want to base it off the prod hotfix branch

PR #1128, opened 2023-01-05
AEA-2837 fix flakey tests in CI build

anthony-nhs:


StefanoFrazzetto (approved):
Great work- thanks again for picking this up!

PR #1129, opened 2023-01-05
AEA-2758 Return invalid prescription

AndyMather24:
https://nhsd-jira.digital.nhs.uk/browse/AEA-2758

When a prescription fails signature verification during releasing prescriptions the failed prescription must be returned to spine - via a DispenseProposalReturn message. 

I've added three interfaces along with the below classes which implement the new interfaces 

- DispenseProposalReturnFactory class responsible for creating DispenseProposalReturns
- DispenseReturnPayloadFactory class responsible for creating the DispenseReturnPayloads
- DispensePropsalReturnHandler class handles the the request and the response logs failure details if required


packages/coordinator/src/services/translation/response/spine-return-handler.ts
```diff
@@ -0,0 +1,45 @@
...
+    returnPayloadFactory: ReturnPayloadFactory,
+    spineClient : SpineClient) {
+    this.requestHeaders = headers
+    this.payloadFactory = returnPayloadFactory
+    this.spineClient = spineClient
+  }
+  async handle(logger: pino.Logger<pino.LoggerOptions>,
+    dispenseProposalReturnRoots: Array<DispenseProposalReturnRoot>) : Promise<void> {
+    const successStatusCodes = 200 || 201
+    Promise.all(
```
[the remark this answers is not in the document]
AndyMather24: good spot, I missed this one I'll need to add some error handling to handle any error from the fire and forget

packages/coordinator/src/services/translation/response/index.ts
```diff
@@ -7,24 +7,44 @@ import {
...
+  ]
+}
+
 export function translateToFhir<T>(
   hl7Message: spine.SpineDirectResponse<T>,
-  logger: pino.Logger): TranslatedSpineResponse {
+  logger: pino.Logger,
+  requestHeaders: Hapi.Util.Dictionary<string>): TranslatedSpineResponse {
+  const releaseHandlers = createReleaseHandlers(requestHeaders)
+  spineResponseHandlers = spineResponseHandlers.concat(releaseHandlers)
```
MatthewPopat-NHS, line 47: this is going to repeatedly add these handlers to the list even if they already exist in there. Should probably leave the spineResponseHandlers as const and just make a local here to store the extra request specific ones.

AndyMather24: silly me, thanks for spotting this!

MatthewPopat-NHS (approved):

MatthewPopat-NHS (approved):

PR #1140, opened 2023-01-09
AEA-2520 Representation of Managing Org Details 

TomSmithCoding:
Altering behaviour for Release response translations, requires some functionality decoupling to handle Healthcare Service resource behaviour.


packages/coordinator/src/services/translation/response/agent-person.ts
```diff
@@ -0,0 +1,149 @@
...
+  locations: Array<fhir.Location>
+  organization?: fhir.Organization
+}
+
+class TranslatedAgentPersonFactory {
+  private readonly agentPerson: hl7V3.AgentPerson
+  private readonly prescriptionType?: string
+  private readonly representedOrganization: hl7V3.Organization
+
+  constructor(agentPerson: hl7V3.AgentPerson, prescriptionType?: string) {
```
AndyMather24, line 22: Hi Tom, looking at the contructor the prescription type can be null. Is there a scenerio where we wouldnt have a prescription type?

packages/coordinator/src/services/translation/response/agent-person.ts
```diff
@@ -0,0 +1,149 @@
...
+    this.prescriptionType = prescriptionType
+    this.representedOrganization = agentPerson.representedOrganization
+  }
+
+  translate(): TranslatedAgentPerson {
+    if (prescriptionRefactorEnabled()) {
+      return this.createWithRefactorEnabled()
+    }
+
+    if (shouldHavePrimaryCareFormat(this.prescriptionType)) {
```
AndyMather24, line 33: making prescription type into and emun will help with readability and adding a switch case

packages/coordinator/src/services/translation/response/common.ts
```diff
@@ -172,130 +162,14 @@ export function convertResourceToBundleEntry(resource: fhir.Resource): fhir.Bund
...
-          name: healthCareOrganization.name
-        }
-      }
-      const organization = createOrganization(hl7Organization)
-      const practitioner = createPractitioner(agentPerson)
-      const practitionerRole = createPractitionerRole(agentPerson, practitioner.id)
-      practitionerRole.organization = fhir.createReference(organization.id)
-      const locations = createLocations(representedOrganization)
-
-      const healthcareService = createHealthcareService(representedOrganization, locations)
```
MatthewPopat-NHS, line 238: we should remove this function, I think this was the only place that called it.

packages/coordinator/tests/services/translation/response/common.spec.ts
```diff
@@ -368,14 +368,12 @@ describe("translateAgentPerson", () => {
 
   test("Prescription is Secondary Care, PrescriptionType starts with 1", () => {
```
MatthewPopat-NHS, line 369: We should probably just remove this test now, it's not really doing what it says anymore

packages/coordinator/tests/services/translation/response/release/release-response.spec.ts
```diff
@@ -251,6 +252,11 @@ describe("bundle resources", () => {
     const organizations = getOrganizations(result)
     expect(organizations).toHaveLength(1)
   })
+  
+  test("does not contain HealthcareService resource", () => {
```
MatthewPopat-NHS, line 256: I would prefer if we just updated our models to never expect healthcare service rather than leaving it as an optional field and testing that we don't populate it.

packages/coordinator/src/services/translation/response/common.ts
TomSmithCoding: Yeah the only other place this appears to be used is in the organization tests

packages/coordinator/tests/services/translation/response/common.spec.ts
TomSmithCoding: Good call

packages/coordinator/src/services/translation/response/agent-person.ts
TomSmithCoding: Not that I'm aware of, there should always be either a prescription type for primary or secondary care.

packages/coordinator/tests/services/translation/response/release/release-response.spec.ts
TomSmithCoding: Yeah I agree that makes more sense, I'll change that to never be expected

packages/coordinator/tests/services/translation/response/organization.spec.ts
```diff
@@ -83,65 +79,21 @@ describe.each([
...
-      expect(fhirHealthcareService.telecom[0].system).toBe("phone")
-      expect(fhirHealthcareService.telecom[0].value).toBe(hl7Organization.telecom._attributes.value.split(":")[1])
-      expect(fhirHealthcareService.telecom[0].use).toBe("work")
-    })
-
-    test("%p has a reference to the location", () => {
-      expect(fhirHealthcareService.location[0].reference).toBe(`urn:uuid:${fhirLocations[0].id}`)
-    })
-
     test("%p has correct address value", () => {
```
MatthewPopat-NHS, line 92: does this still work? I thought the location resources were only created in the secondary care format.

TomSmithCoding: It does still pass yeah, and the createLocations appears to still be used within the `createWithRefactorEnabled` function

packages/coordinator/src/services/translation/response/agent-person.ts
```diff
@@ -0,0 +1,149 @@
...
+import {createPractitioner} from "./practitioner"
+import {createLocations, createOrganization, getOrganizationCodeIdentifier} from "./organization"
+import {createPractitionerRole, createRefactoredPractitionerRole} from "./practitioner-role"
+import {createPractitionerOrRoleIdentifier} from "./identifiers"
+import {prescriptionRefactorEnabled} from "../../../utils/feature-flags"
+import {addIdentifierToPractitionerOrRole} from "./common"
+
+interface TranslatedAgentPerson {
+  practitionerRole: fhir.PractitionerRole
+  practitioner?: fhir.Practitioner
```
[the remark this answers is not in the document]
TomSmithCoding: It appears to be nullable because of the `createWithRefactorEnabled` function.

AndyMather24 (approved):

PR #1169, opened 2023-01-16
AEA-2775 identify number of repeats

AndyMather24:
- Added tests to confirm repeat number map to prescription and line items
- lows and highs mapped as per the AC's  with tests [https://nhsd-jira.digital.nhs.uk/browse/AEA-2775](ticket)


yorkshiremoose (approved):
Ship it.

PR #1179, opened 2023-01-17
Aea 2957 update lossless json

anthony-nhs:


packages/coordinator/src/services/translation/response/release/release-medication-request.ts
```diff
@@ -128,15 +128,15 @@ function createBasedOn(
     url: "https://fhir.nhs.uk/StructureDefinition/Extension-EPS-RepeatInformation",
     extension: [{
       url: "numberOfRepeatsAllowed",
-      valueInteger: new LosslessNumber(parseInt(prescriptionRepeatNumber.high._attributes.value))
+      valueInteger: new LosslessNumber(parseInt(prescriptionRepeatNumber.high._attributes.value).toString())
```
MatthewPopat-NHS, line 131: This seems wrong, surely if we want a lossless number we should use the original value rather than parsing and immediately stringifying.

MatthewPopat-NHS (approved):

PR #1204, opened 2023-01-25
AEA-2661 Dispensing org within release rejection

originalphil:
…eExtension containing org id, instead of ODS code.

### DIFF
--- a/packages/coordinator/src/services/translation/response/spine-response-handler.ts
+++ b/packages/coordinator/src/services/translation/response/spine-response-handler.ts
@@ -575,19 +575,14 @@ export class ReleaseRejectionHandler extends SpineResponseHandler<hl7V3.Prescrip
     const spineResponse = super.handleErrorResponse(sendMessagePayload, logger)
     const operationOutcome = spineResponse.fhirResponse as fhir.OperationOutcome
 
-    let withAnotherDispenser = false
-    operationOutcome.issue.forEach((issue) => {
-      if (ReleaseRejectionHandler.withAnotherDispenser(issue)) {
-        withAnotherDispenser = true
-        issue.diagnostics = ReleaseRejectionHandler.getDiagnosticInfo(sendMessagePayload)
-      }
-    })
-    if (withAnotherDispenser) {
+    if (operationOutcome.issue.some(
+      (issue) => ReleaseRejectionHandler.withAnotherDispenser(issue)
+    )) {
       const organization = ReleaseRejectionHandler.getOrganizationInfo(sendMessagePayload)
       operationOutcome.contained = [organization]
-      const extension: fhir.IdentifierReferenceExtension<fhir.Bundle> = {
+      const extension: fhir.ReferenceExtension<fhir.Bundle> = {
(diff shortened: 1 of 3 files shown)


omarisgreat:
Look like this addresses the issues raised on the ticket 👍

packages/coordinator/src/services/translation/response/spine-response-handler.ts
```diff
@@ -575,19 +575,14 @@ export class ReleaseRejectionHandler extends SpineResponseHandler<hl7V3.Prescrip
...
+    if (operationOutcome.issue.some(
+      (issue) => ReleaseRejectionHandler.withAnotherDispenser(issue)
+    )) {
       const organization = ReleaseRejectionHandler.getOrganizationInfo(sendMessagePayload)
       operationOutcome.contained = [organization]
-      const extension: fhir.IdentifierReferenceExtension<fhir.Bundle> = {
+      const extension: fhir.ReferenceExtension<fhir.Bundle> = {
         url: "https://fhir.nhs.uk/StructureDefinition/Extension-Spine-supportingInfo",
-        valueReference: {identifier: organization.identifier[0]}
+        valueReference: {reference: `#${organization.id}`}
```
StefanoFrazzetto, line 585: Should the prefix for the reference be this `urn:uuid:<RESOURCE_UUID>`, instead of `#<RESOURCE_UUID>`?

> When a contained resource references its container or another contained resource within its container, it must always be a local relative reference (i.e. it must be or start with '#'). It's never possible to point to the container or a contained resource within it by absolute or versioned reference
https://build.fhir.org/references.html

StefanoFrazzetto (commented):
I spotted one minor potential issue, but the rest look good!

packages/coordinator/src/services/translation/response/spine-response-handler.ts
omarisgreat: @StefanoFrazzetto, I believe the quote you've given reinforces that Phil has done this correctly. The supportingInfo reference is pointing to a contained resource, so the resource ID must start with '#'

StefanoFrazzetto: @omarisgreat I thought that mean you're supposed to reference other resources from within the same contained block using the '#'. To avoid any confusion in the future, since I'm still not too familiar with FHIR docs, would you be able to point us to some examples?

omarisgreat: @StefanoFrazzetto, the # is used to reference any contained resource, either within the contained block or from the containing resource. see snippets on https://hl7.org/fhir/r4/references.html#contained, alternatively, we have a few examples of these in our dispense notification messages, e.g. https://simplifier.net/guide/NHSDigital-Medicines/Home/Examples/AllExamples/Messagedispense-notification/prescription-dispense-notification.guide.md?version=current (references to performer and request are outside the contained block)

StefanoFrazzetto (approved):

TomSmithCoding (approved):

PR #1280, opened 2023-02-08
AEA-3029 Run e2e pact tests in a test runner

kris-szlapa:


kris-szlapa:
The ticket AEA-3029 has been rejected

PR #1298, opened 2023-02-15
AEA-3043 amend dispense notification details in epsat

JackSpagnoliNHS:


packages/tool/site/client/src/pages/dispensePage.tsx
```diff
@@ -213,4 +213,8 @@ function containsParacetamol(medicationRequest: fhir.MedicationRequest): boolean
   return medicationRequest.medicationCodeableConcept.coding[0].code === ("39720311000001101")
 }
 
+export function shouldSendCustomFhirRequest(dispenseFormValues: DispenseFormValues) {
```
MatthewPopat-NHS, line 216: seems unnecessary to put this function here, might as well just check this condition inline in createDispenseNotification.ts

PR #1322, opened 2023-02-22
AEA-3074 Provide support for multiple order intents within the API

originalphil:


packages/coordinator/tests/services/validation/bundle-validator.spec.ts
```diff
@@ -267,6 +271,126 @@ describe("verifyCommonBundle", () => {
...
+        fhir.MedicationRequestIntent.ORDER,
+        fhir.MedicationRequestIntent.ORIGINAL_ORDER
+      )
+      const validationErrors = validator.verifyCommonBundle(bundle, "test_sds_user_id", "test_sds_role_id")
+
+      expect(validationErrors).toHaveLength(1)
+      // eslint-disable-next-line max-len
+      expect(validationErrors[0].diagnostics).toEqual("MedicationRequest.intent must be original-order or instance-order.")
+    })
+  })
```
StefanoFrazzetto, line 391: Consider reducing duplication by parametrising the tests, e.g.

```ts
test.each([
  ["order", [fhir.MedicationRequestIntent.ORDER]],
  ["original-order/instance-order", [fhir.MedicationRequestIntent.ORIGINAL_ORDER, fhir.MedicationRequestIntent.INSTANCE_ORDER]]
])("Should return empty list given all MedicationRequest intents are %s", (
  description: string,
  medicationRequestIntents: Array<REPLACE_WITH_MEDICATION_INTENT_TYPE>
) => {

  setMedicationRequestIntents(bundle, ...medicationRequestIntents)

  const validationErrors = validator.verifyCommonBundle(bundle, "test_sds_user_id", "test_sds_role_id")
  expect(validationErrors).toHaveLength(0)
})
```

packages/coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -158,6 +154,46 @@ export function verifyCommonBundle(
...
+): Array<fhir.OperationOutcomeIssue> {
+  const incorrectValueErrors: Array<fhir.OperationOutcomeIssue> = []
+
+  const repeatPrescribing = isRepeatPrescribing(medicationRequests)
+  const repeatDispensing = isRepeatDispensing(medicationRequests)
+  const acute = !(repeatPrescribing || repeatDispensing)
+
+  if (acute && medicationRequests.some(
+    medicationRequest => medicationRequest.intent !== fhir.MedicationRequestIntent.ORDER
+  )) {
```
StefanoFrazzetto, line 168: Instead of repeating `medicationRequests.some(...)`, could you do something to make the if statements more readable? For example, something like this would avoid the repetitions:

```ts
const intentDoesNotMatch = (
  medicationRequests: fhir.MedicationRequest[],
  intent: fhir.MedicationRequestIntent
): boolean => {
  return medicationRequests.some(medicationRequest => medicationRequest.intent !== intent)
}

if (acute && intentDoesNotMatch(medicationRequests, fhir.MedicationRequestIntent.ORDER) {
  ...
}
```

StefanoFrazzetto (changes requested):
Nice work. As you mentioned, there are duplicated code blocks, but it should be feasible to create functions, as well as parametrising tests, to reduce the number of repetitions.

MatthewPopat-NHS (approved):

PR #1337, opened 2023-02-24
AEA-2963 Add pre-commit to handle git hooks

StefanoFrazzetto:
Reduces the number of unnecessary pre-commit checks we do, and selectively test changes in components.


originalphil (approved):

StefanoFrazzetto:
Renamed the PR, as the changes here match one of the items of AEA-2963.

PR #1363, opened 2023-03-06
AEA-3110 Resolve SonarCloud Findings

JackSpagnoliNHS:


packages/coordinator/src/services/communication/sandbox-spine-client.ts
```diff
@@ -32,15 +29,14 @@ export class SandboxSpineClient implements SpineClient {
...
-      statusCode: 400,
-      body: notSupportedOperationOutcome
-    })
+    return notSupportedOperationOutcomePromise()
   }
 
   async handleSpineRequest(spineRequest: spine.SpineRequest): Promise<spine.SpineResponse<unknown>> {
     switch (spineRequest.interactionId) {
       case hl7V3.Hl7InteractionIdentifier.PARENT_PRESCRIPTION_URGENT._attributes.extension:
+      case hl7V3.Hl7InteractionIdentifier.DISPENSE_NOTIFICATION._attributes.extension:
```
AndyMather24, line 38: This looks like we are changing the intent of the logic - we need to check this is correct as per business rules

packages/coordinator/src/services/communication/tracker/tracker-response-builder.ts
```diff
@@ -21,9 +21,12 @@ interface TrackerResponse {
 }
 
 const createTrackerError = (error: TrackerErrorString, details?: string | Array<string>): TrackerError => {
```
AndyMather24, line 23: Feel free to test this but I'm sure we can write the function like this which makes it a little more readable `const createTrackerError = (error: TrackerErrorString, details?: string | Array<string>): TrackerError => {
  let messageDetails : Array<string>

  if(Array.isArray(details)) {
    messageDetails = details
  }
    
  return {
    errorMessage: error,
    errorDetails: messageDetails ? messageDetails : []
  }
}`

packages/coordinator/src/services/translation/request/dosage.ts
```diff
@@ -194,33 +194,19 @@ function stringifyFrequencyAndPeriod(dosage: fhir.Dosage) {
   const periodMax = repeat?.periodMax
   const periodUnit = repeat?.periodUnit
 
-  if (!frequency && !frequencyMax) {
```
AndyMather24, line 197: This looks like quite a tricky bit of code to refactor. Whats the tests like around this code are we confident we understand what the code was/should be doing? 

If we dont we can look at the git history and look at the ticket for the original code to ensure we nothing is broken

AndyMather24 (commented):
Looks like the complaint was regarding the assignment on line 193 as the  sortIndex = 0 is already asigned on line 190.

packages/coordinator/src/services/translation/response/common.ts
```diff
@@ -187,37 +187,27 @@ export function orderBundleResources(r1: fhir.Resource, r2: fhir.Resource): numb
 }
 
 function getSortIndex(resource: fhir.Resource) {
-  let sortIndex = 0
   switch (resource.resourceType) {
     case "MessageHeader":
-      sortIndex = 0
```
AndyMather24, line 193: Looks like the complaint was for assigning 0 to sortIndex after already assigning it on line 190

packages/e2e-tests/auth/src/oauth.ts
```diff
@@ -58,7 +58,7 @@ export class AuthClient {
     this.environment = this.getEnvironment()
     this.callbackUrl = "https://example.org/" // using /callback causes the website to return a 404, which upsets axios
     this.state = this.getState()
-    this.scope = ""
+    //this.scope = ""
```
AndyMather24, line 61: If we dont need this lets remove it

packages/coordinator/src/services/translation/request/dosage.ts
AndyMather24, line 197: After looking at the refactor I'm confident its like for like.

An approach to take in future when refacting is to ensure there are tests for func we are refactoring. If no tests then write some and see them pass then do the refactoring and you'll know if anything has been broken with the refactor

AndyMather24: After looking at the refactor I'm confident its like for like.

An approach to take in future when refactor is to ensure there are tests for func we are refactoring. If no tests then write some and see them pass then do the refactoring and you'll know if anything has been broken with the refactor

AndyMather24 (approved):
Good work! 

Although would be good to get someone more familiar with python to review .py files

packages/coordinator/src/services/translation/request/dosage.ts
```diff
@@ -230,6 +216,48 @@ function stringifyFrequencyAndPeriod(dosage: fhir.Dosage) {
...
+  period: string | LosslessNumber,
+  periodMax: string | LosslessNumber,
+  periodUnit: fhir.UnitOfTime
+){
+  if (!period && !periodMax) {
+    return []
+  } else if (isOne(period) && !periodMax) {
+    return [getReciprocalUnitOfTimeDisplay(periodUnit)]
+  } else {
+    //TODO - why is this fine when period is 1 but not otherwise?
```
StefanoFrazzetto, line 230: Check TODO

StefanoFrazzetto (commented):
Good work- I have only checked the Python changes. My Python is probably a bit rusty, but I've attempted refactoring `update_prescriptions.py`, splitting it into modules and moving tests to a separate location, so that hopefully the core logic should be easier to follow.

Please have a look and see if everything makes sense. I'll give the PR another look tomorrow to make sure I haven't missed anything.

StefanoFrazzetto (approved):

PR #1385, opened 2023-03-08
Made SDS-role-profile-id optional

MatthewPopat-NHS:


MatthewPopat-NHS:
This didn't end up being needed

PR #1400, opened 2023-03-15
AEA-2984-Accept-both-message-formats

TomSmithCoding:


anthony-nhs (changes requested):
code changes are good but could just do with some more testing.   

As well as the unit test change I have pointed out, all current examples under examples should be changed to remove practionerRole.healthCareService (as that should be the default) but there should be an example added to ensure everything works when practionerRole.healthCareService is still passed in (to ensure we do not break anything with IC24)

packages/coordinator/src/services/validation/bundle-validator.ts
```diff
@@ -453,13 +453,8 @@ function checkPrimaryCarePrescriptionResources(
...
 
 function checkSecondaryCarePrescriptionResources(
-  practitionerRole: fhir.PractitionerRole,
   organization: fhir.Organization,
 ): fhir.OperationOutcomeIssue {
-  if (!practitionerRole.healthcareService) {
-    return errors.missingRequiredField("practitionerRole.healthcareService")
-  }
-
   if (organization.partOf) {
```
MatthewPopat-NHS, line 458: this should be allowed for secondary care

packages/specification/schemas/MedicationRequest/extensions/NHSBSAPrescriptionType.yaml
```diff
@@ -5,8 +5,6 @@ description: |
   For the full list of prescription types, and what they mean, please refer to the following [guidance](https://www.nhsbsa.nhs.uk/sites/default/files/2022-07/NHSBSAGuidanceforEndorsement_v7.7.pdf) from the BSA.
   
   Where the prescription type is of the pattern 01nn, the details of the prescribing organisation must be captured within the Organisation resource.
```
MatthewPopat-NHS, line 7: We should probably remove this line too

examples/errors/1-Process-Request-Send-400_INVALID_CHECKSUM.json
```diff
@@ -358,40 +352,6 @@
         ]
       }
     },
-    {
-      "fullUrl": "urn:uuid:54b0506d-49af-4245-9d40-d7d64902055e",
```
MatthewPopat-NHS, line 362: Unfortunately for all these examples we probably shouldn't be just removing the Healthcare service resource, we should be translating it to the equivalent organisation resource.

packages/coordinator/src/services/validation/bundle-validator.ts
TomSmithCoding: What should be allowed sorry? We are now allowing healthcareService resources for Secondary Care prescriptions, or are we saying they should be allowed to include the `organization.partOf` to replace the healthcareService resource (I think I just answered my own question)

packages/specification/electronic-prescription-service-api.yaml
```diff
@@ -975,7 +975,6 @@ components:
             - $ref: "#/components/schemas/Create-Practitioner-Resource"
             - $ref: "#/components/schemas/Create-MessageHeader-Resource"
             - $ref: "#/components/schemas/Create-Patient-Resource"
-            - $ref: "#/components/schemas/Create-HealthcareService-Resource"
             - $ref: "#/components/schemas/Create-Location-Resource"
```
MatthewPopat-NHS, line 978: We should remove the Location resources as well

MatthewPopat-NHS (approved):

PR #1408, opened 2023-03-20
publish code coverage

anthony-nhs:


packages/tool/site/server/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "EPS API Tool Server",
   "main": "app.js",
   "scripts": {
-    "test": "jest --ci --reporters=default --reporters=jest-junit",
+    "test": "jest --ci --reporters=default --reporters=jest-junit --coverage  --coverageReporters=cobertura --coverageReporters=text --coverageReporters=html",
```
originalphil, line 7: I don't suppose these are still needed, if we're only publishing the client coverage?

anthony-nhs: good point - removed

originalphil (approved):

PR #1425, opened 2023-03-30
AEA-2934

ayoseg:


packages/bdd-tests/step_definitions/shared-steps.ts
```diff
@@ -0,0 +1,113 @@
+import { getToken } from "../services/getaccessToken";
```
MatthewPopat-NHS, line 1: It would make more sense to define these inside the other files in this folder. authentication is the only one that I would say is actually shared.

packages/bdd-tests/step_definitions/releaseprescription-steps.ts
```diff
@@ -0,0 +1,97 @@
...
+        .toBe("Paracetamol 500mg soluble tablets")
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[2].resource.dispenseRequest.quantity.value).toEqual(60)
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[3].resource.medicationCodeableConcept.coding[0].display)
+        .toBe("Methotrexate 10mg/0.2ml solution for injection pre-filled syringes")
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[3].resource.dispenseRequest.quantity.value).toEqual(1)
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[4].resource.medicationCodeableConcept.coding[0].display)
+        .toBe("Flucloxacillin 500mg capsules")
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[4].resource.dispenseRequest.quantity.value).toEqual(28)
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[_number+1].resource.resourceType).toBe("Patient")
+      expect(ss.resp.data.parameter[1].resource.entry[1].resource.entry[_number+1].resource.identifier[0].value).toBe("9449304130")
```
MatthewPopat-NHS, line 68: we probably shouldn't assume that the medication request will always be the first entry in the bundle

packages/bdd-tests/step_definitions/dispensenotification-steps.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    test('Send a dispense notification for an acute prescription', ({ given, and, when, then }) => {
+
+      ss.givenIAmAuthenticated(given)
+
+      ss.givenICreateXPrescriptionsForSite(given)
+
+      ss.whenIReleaseThePrescription(when)
+
+    and('the prescription status is With Dispenser', async() => {
```
MatthewPopat-NHS, line 21: If we're going to leave these intentionally unimplemented for now we should  leave TODO comments in here to indicate as such.

packages/bdd-tests/src/configs/ap.ts
```diff
@@ -0,0 +1,39 @@
+//import * as crypto from "crypto";
```
MatthewPopat-NHS, line 1: not sure what this file is, I'm guessing you didn't mean to commit it.

packages/bdd-tests/src/configs/api.ts
```diff
@@ -0,0 +1,54 @@
...
+
+let url = process.env.base_url;
+let dir = './resources';
+
+let instance = axios.create({
+  baseURL: url,
+  headers: {
+    "NHSD-Session-URID": "555254242106",
+    "X-Request-ID": genid.generateRandomUUID(),
+    "X-Correlation-ID": "11C46F5F-CDEF-4865-94B2-0EE0EDCC26DA",
```
MatthewPopat-NHS, line 13: This shouldn't be hard coded, it should either be generated as above or not provided

packages/bdd-tests/features/releaseprescription.feature
```diff
@@ -0,0 +1,39 @@
...
+  Scenario: Release a prescription with multiple line item for a dispensing site
+    Given I create 1 prescription(s) for FGG90 with 4 line items
+    When I release the prescriptions
+    Then I get 1 prescription(s) released to FGG90
+    And 4 line items are returned in the response
+
+  @excluded
+  Scenario: Release a prescription with over 4 line items for a dispensing site - invalid
+    Given I create 1 prescription(s) for FGG90 with 5 line items
+    When I release the prescriptions
```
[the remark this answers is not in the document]
ayoseg: I agree I will move this to create feature.

packages/bdd-tests/npm
MatthewPopat-NHS, line 1: These empty files don't seem right

packages/bdd-tests/util/helpercont.ts
```diff
@@ -0,0 +1,83 @@
+import * as crypto from "crypto";
```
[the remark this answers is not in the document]
ayoseg: Ok, I did this so as not to make the helper file too big. But as I don't have a better name,  I will combine the files for clarity purposes

packages/bdd-tests/step_definitions/shared-steps.ts
ayoseg: You can't have the whole function inside the shared steps. It is not clean to have the get access Token steps in the shared step. Shared steps is ideal for putting the getToken call, which I am doing. But the whole logic should be put in a different file. May I am missing your point here

packages/bdd-tests/step_definitions/releaseprescription-steps.ts
ayoseg: I haven't yet seen an example where it isn't. As we haven't sorted out the signature and the response getting in the wrong area, I am happy to leave this as it is now. It can be revisited when we get the response in right area, ie the passedPrescription section in the Json. Explain what you mean by dedup of lookup please

packages/bdd-tests/step_definitions/dispensenotification-steps.ts
ayoseg: Unfortunately as we don't have the tracker - which is why we have the unimplemented, the code will just be littered with todo, best to get push on getting that resolved. I could put todo there anyway

packages/bdd-tests/src/configs/ap.ts
ayoseg: has been taken off in the latest commit. Wasn't meant to be there

packages/bdd-tests/src/configs/api.ts
ayoseg: Agreed

packages/bdd-tests/npm
ayoseg: Adding to git ignore

packages/bdd-tests/services/getaccessToken.ts
```diff
@@ -0,0 +1,72 @@
...
+
+axios.interceptors.response.use(response => {
+  return response;
+}, error => {
+  console.log(`Environment is ${process.env.NODE_ENV}`)
+  console.log(`Status code ${error.response.status} : Message - ${error.response.statusText}`)
+  console.log(error.response.data)
+  console.log(JSON.stringify(error.response.data.issue[0].details))
+  return Promise.reject(error);
+});
```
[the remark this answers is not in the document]
ayoseg: This above is used specifically on Requests to get Access token. The other one is used on requests to EPS. these are two different services.

packages/bdd-tests/services/getaccessToken.ts
```diff
@@ -0,0 +1,72 @@
...
+    resp = await Req().adhocPost(`${process.env.base_url}/oauth2-mock/token`, qs.stringify({
+      grant_type: "authorization_code",
+      client_id: process.env.client_id,
+      client_secret: process.env.client_secret,
+      redirect_uri: process.env.redirect_uri,
+      code: code
+    }),
+      { headers: {'Content-Type': 'application/x-www-form-urlencoded'}})
+    console.log("======= Access Token ======== " + resp.data.access_token)
+  return resp.data.access_token
```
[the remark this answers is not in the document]
ayoseg: I agree, will refactor

MatthewPopat-NHS:
this got merged in with https://github.com/NHSDigital/electronic-prescription-service-api/pull/1432

PR #1434, opened 2023-04-05
AEA-3016-Release-Rejection-Message-FHIR-Standards

TomSmithCoding:


originalphil (commented):
Hey Tom. You seem to have added a convertMomentToISODateTime function, when that already exists in common/dateTime.
I also can't see where you've then made use of that function in your changes. I can only see the existing usage of the existing function.

TomSmithCoding:
> Hey Tom. You seem to have added a convertMomentToISODateTime function, when that already exists in common/dateTime. I also can't see where you've then made use of that function in your changes. I can only see the existing usage of the existing function.

Hey Phil cheers for looking over it. For the convertMoment function I was having trouble referencing the existing function from common/dateTime, the import path didn't seem to work. Would you happen to know what the path should be?
As for the usage of that function, per my slack message I initially attempted to use it on line 13 in the lastUpdated field however this caused some issues within the pact tests and the examples, so I changed it back to the hardcoded example date for now and looking for further input.

originalphil (approved):

PR #1446, opened 2023-04-19
AEA-3000 Allow Prepare to Generate SHA2 Digests

JackSpagnoliNHS:


packages/coordinator/src/routes/create-hash.ts
```diff
@@ -0,0 +1,12 @@
+import * as crypto from "crypto-js"
+import {HashingAlgorithm} from "../../src/services/translation/common/hashingAlgorithm"
+
+type Encoder = typeof crypto.enc.Base64
+export function createHash(thingsToHash: string, hashingAlgorithm: HashingAlgorithm, encoder?: Encoder): string {
```
originalphil, line 5: Default value of SHA1 on the hashingAlgorithm?

packages/coordinator/tests/logging/expectations.ts
```diff
@@ -3,20 +3,30 @@ import Hapi from "@hapi/hapi"
...
+
+const SHA1_HASH_LENGTH = 40
+const SHA256_HASH_LENGTH = 64
 
 /**
  * Expects that the hash for incoming payloads is logged.
  * @param logs - the logs produced for a request to the API
  */
-const expectPayloadAuditLogs = (logs: Array<Hapi.RequestLog>): void => {
+const expectPayloadAuditLogs = (logs: Array<Hapi.RequestLog>, path: string, expectedUseSHA256?: boolean): void => {
```
originalphil, line 15: Another default value might be useful here. Then no need for the ?? operator further along. I think there was another example of this in another file that I've collapsed so now can't find...

MatthewPopat-NHS (approved):

PR #1485, opened 2023-05-12
AEA-2803 Claim Endpoint - Response Schema

taiyeoguns:


AmeyaKrishnamoorthy (approved):
Looks good to me. I'd just double check the references actually resolve as this has been an issue in the past; otherwise approved.

PR #1488, opened 2023-05-15
AEA-3230 Create GitHub self-hosted runners and pipelines

originalphil:


MatthewPopat-NHS:
This has now been moved over to a different repo

PR #1489, opened 2023-05-16
AEA-2786-Withdraw-Request-Schema

TomSmithCoding:


packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+        type: string
+        example: "#16708936-6397-4e03-b84f-4aaa790633e0"
+  reasonCode:
+    type: object
+    description: The reason for the task, should only be included if there is no focus or differs from the reason indicated in the focus.
+    properties:
+      coding:
+        type: array
+        items:
+          $ref: ../../components/_fragments/codes.yaml#/components/schemas/ReasonCode
```
MatthewPopat-NHS, line 90: We shouldn't use a standard import here, we should be indicating the valid list of reason codes in here.

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+    properties:
+      identifier:
+        $ref: ../../components/_fragments/identifiers.yaml#/components/schemas/OdsOrganizationCode
+  requester:
+    type: object
+    description: The dispenser withdrawing the prescription.
+    properties: 
+      reference:
+        type: string
+        example: "#16708936-6397-4e03-b84f-4aaa790633e0"
```
MatthewPopat-NHS, line 82: This doesn't look valid, I would copy the example and description from the equivalent field in MedicationRequest.yaml

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+      identifier:
+        $ref: ../../components/_fragments/identifiers.yaml#/components/schemas/Rfc4122
+  for:
+    type: object
+    properties: 
+      identifier:
+        $ref: ../../components/_fragments/identifiers.yaml#/components/schemas/NhsNumber
+  authoredOn:
+    type: string
+    description: The time the prescription is being withdrawn to Spine.
```
MatthewPopat-NHS, line 67: "withdrawn *from*"

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+      coding:
+        type: array
+        items:
+          $ref: ../../components/_fragments/codes.yaml#/components/schemas/TaskCode
+  groupIdentifier:
+    description: The Prescription ID of the prescription that is being withdrawn.
+    $ref: ../../components/_fragments/groupIdentifier.yaml
+  focus:
+    type: object
+    description: The release response which released the prescription that is currently being withdrawn.
```
MatthewPopat-NHS, line 56: it's DN identifier, not the prescription

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+    example: order
+  code: 
+    type: object
+    properties:
+      coding:
+        type: array
+        items:
+          $ref: ../../components/_fragments/codes.yaml#/components/schemas/TaskCode
+  groupIdentifier:
+    description: The Prescription ID of the prescription that is being withdrawn.
```
MatthewPopat-NHS, line 52: "The prescription id that the DN was originally submitted against.

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+    properties:
+      coding: 
+        type: array
+        items:
+          $ref: ../../components/_fragments/codes.yaml#/components/schemas/EpsTaskDispenseWithdrawStatusReason
+  intent:
+    type: string
+    description: The intent of the Medication Requests being withdrawn.
+    example: order
+  code:
```
MatthewPopat-NHS, line 44: Not sure what this is, can't see it being used in our code.

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+    items:
+      $ref: ../../components/_extensions/EPSRepeatInformation.yaml
+  identifier:
+    description: The unique identifier for the Withdraw Request.
+    type: array
+    items:
+      $ref: ../../components/_fragments/identifiers.yaml#/components/schemas/Rfc4122
+  status: 
+    type: string
+    example: in-progress
```
MatthewPopat-NHS, line 31: this is the only value that's valid for a withdraw not just an example, so should be enum

packages/specification/schemas/requests/dispensing/Withdraw.yaml
```diff
@@ -0,0 +1,90 @@
...
+      oneOf:
+      - $ref: ../../components/PractitionerRole/TaskPractitionerRole.yaml
+      - $ref: ../../components/Organization/DispenseNotificationOrganization.yaml
+  extension:
+    type: array
+    description: |
+      FHIR extension array to hold Repeat Information.
+      This identifies the instance of the eRD prescription that is being returned.
+      Given that all eRD prescriptions have the same prescription ID,
+      the combination of Prescription ID and repeat number identifies the issue to withdraw or return.
```
MatthewPopat-NHS, line 21: "... identifies the issue *that the DN was originally submitted against.*"

packages/specification/schemas/requests/dispensing/Withdraw.yaml
TomSmithCoding: Task-code I can currently see being used in task-validator.ts and translation.ts, unless you're referring to the Intent which I can only see being forced to TaskIntent.ORDER in translation, and anything else fails the task-validator

MatthewPopat-NHS: Ah yes, again we should probably inline this and explicitly state that it must be "abort".

TomSmithCoding: Will do

originalphil (approved):

PR #1506, opened 2023-06-07
AEA-2882 fix BDD tests

anthony-nhs:


packages/bdd-tests/services/getaccessToken.ts
```diff
@@ -1,64 +1,257 @@
...
+  getKeycloakUrl(): string {
+    return `https://identity.ptl.api.platform.nhs.uk/auth/realms/Cis2-mock-internal-dev/protocol/openid-connect`
   }
-};
 
-const cookieJar = {
-  myCookies: undefined,
-};
+  private getState(): string {
+    // return uuid.v4()
```
originalphil, line 66: Does this want un-commenting and the below line removing?

Also, this file should probably be called getAccessToken.ts

packages/bdd-tests/services/getaccessToken.ts
```diff
@@ -1,64 +1,257 @@
...
+ *
+ * 4. The mock auth redirected us back to the
+ * identity-service, which redirected us to whatever our app's
+ * callback-url was set to.  We don't actually care about the
+ * content our callback-url page, we just need the auth_code that
+ * was provided in the redirect.
+ *
+ * 5. Finally, get an access token.
+ *
+ * 6. Profit... sweet sweet profit.
```
originalphil, line 25: Nice :D

packages/bdd-tests/step_definitions/createPrescriptionSteps.ts
```diff
@@ -0,0 +1,21 @@
+import {Then} from "@cucumber/cucumber"
+
+Then(/^IGNORE I get an error response (\d+)$/, function (status, table) {
```
originalphil, line 3: Do these not clash? How does cucumber choose which one to invoke?

packages/bdd-tests/step_definitions/dispenseNotificationSteps.ts
```diff
@@ -0,0 +1,15 @@
+import * as helper from "../util/helper"
+
+import {When} from "@cucumber/cucumber"
+
+When(/^I amend the dispense notification for item (\d+)$/, async function (itemNo, table) {
+  this.resp = await helper.amendDispenseNotification(itemNo, table, this)
+})
+
+When("I withdraw the dispense notification", async function (table) {
```
originalphil, line 9: More clashing? Maybe I'm not understanding cucumber properly?

packages/bdd-tests/util/helper.ts
```diff
@@ -1,67 +1,73 @@
 import * as crypto from "crypto"
```
originalphil, line 1: This file probably wants carving up into some more descriptive ones. A job for another time though.

packages/bdd-tests/services/getaccessToken.ts
anthony-nhs: renamed file
The file is just a copy of packages/e2e-tests/auth/src/oauth.ts so that included the commented out line as well

packages/bdd-tests/step_definitions/dispenseNotificationSteps.ts
anthony-nhs: You are right - there is a clash.
The tests were originally written using cucumber-jest framework which allowed different definitions for different tests. 
This now wont work, but the tests for dispense do not currently run so it does not show as a clash

packages/bdd-tests/step_definitions/createPrescriptionSteps.ts
anthony-nhs: You are right - there is a clash.
The tests were originally written using cucumber-jest framework which allowed different definitions for different tests.
This now wont work, but the tests for dispense do not currently run so it does not show as a clash

packages/bdd-tests/util/helper.ts
anthony-nhs: yeah - because it originally used local variables to manage state it all needed to be in one file.  As it uses context to store state now it can be in multiple files
Have added a comment saying it should be split to smaller files

originalphil (approved):

PR #1562, opened 2023-09-05
AEA-3204 Switched to new mock auth

MatthewPopat-NHS:


TomSmithCoding (approved):
Agreed to merge with what we've done so far, and open new ticket to fix selenium tests.

PR #1875, opened 2024-01-05
AEA-3153-Allow-ResponsibleParty-to-be-Org

TomSmithCoding:
…onal

## Summary

Validating that prescriber endpoint will accept a PractitionerRole without Practitioner, Identifier or Code references.
Updating spec to reflect these values as optional
Provide an example of prescription which contains an org as ResponsibleParty
 - Responsible practitioner org should be the name of one of the new example (displayed in the drop down) 

## Reviews Required

- [x] Dev
- [ ] Test
- [ ] Tech Author

## Review Checklist

- [ ] I have reviewed the changes in this PR and they fill all or part of the acceptance criteria of the ticket, and the code is in a mergeable state.
- [ ] If there were infrastructure, operational, or build changes, I have made sure there is sufficient evidence that the changes will work.
- [ ] I have ensured the jira ticket has been updated with the github pull request link


packages/specification/electronic-prescription-service-api.yaml
```diff
@@ -315,6 +315,11 @@ paths:
                 description: a Valid eRD, primary care prepare request.
                 value:
                   $ref: examples/spec/prepare/eRD-primary-example-req.json
+              responsible-party-org:
+                summary: Responsible practitioner org 
+                description: a Valid example with an organisation as the responsible practitioner, primary care prepare request.
+                value:
+                  $ref: examples/spec/prepare/responsible-party-org-example-req.json
```
MatthewPopat-NHS, line 322: these should be using the same files as the pact tests

examples/spec/prepare/responsible-party-org-example-req.json
MatthewPopat-NHS, line 1: shouldn't have these duplicate files

packages/coordinator/src/services/translation/response/cancellation/cancellation-response.ts
```diff
@@ -63,7 +63,7 @@ function createBundleEntries(cancellationResponse: hl7V3.CancellationResponse) {
   //The ResponsibleParty represents the author of the original prescription (if different to the cancel requester)
   const originalPrescriptionAuthorAgentPerson = cancellationResponse.responsibleParty?.AgentPerson
   let translatedOriginalPrescriptionAuthor = translatedCancelRequester
-  if (originalPrescriptionAuthorAgentPerson) {
+  if (originalPrescriptionAuthorAgentPerson.agentPerson) {
```
MatthewPopat-NHS, line 66: This is now assuming that the responsible party is present which I don't think is always true

MatthewPopat-NHS (approved):

PR #2028, opened 2024-02-13
AEA2962 - Provide a clear error message when the ODS code of the reimbursement authority is missing

natasafrgk:
## Summary

For a dispense notification, there should be a check in place to look for an ODS code value for a reimbursement authority in Organization. If this value doesn't exist then an error should be thrown to inform about the absence of the value.

Then the error should be added in the EPS API documentation under "Validation Errors", with the additional mention that the ODS code should be provided. 

https://nhsd-jira.digital.nhs.uk/browse/AEA-2962

- Routine Change

### Details

The changes have happened in the dispense-notification and documentation. Logic for checking the BSAid , which refers to the value of the ODS code for the reimbursed authority was added, along with the respective unit test.

Then, to increase the cove coverage of the file, another unit test was added to check if an error was thrown when the organisationRef wasn't a reference. This happened as part of a generic tidying up.


packages/coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
```diff
@@ -243,6 +243,26 @@ describe("fhir MedicationDispense maps correct values in DispenseNotification wh
...
+    try {
+      convertDispenseNotification(dispenseNotification, logger)
+    } catch (e) {
+      expect(e.userErrorCode).toEqual("INVALID_VALUE")
+      expect(e.userErrorMessage).toEqual(
+        // eslint-disable-next-line max-len
+        "The dispense notification is missing the ODS code for the reimbursed authority."
+      )
+      expect(e.userErrorFhirPath).toEqual("Organization.extension[0].extension[0].valueIdentifier.value")
+    }
```
JackSpagnoliNHS, line 262: Could we use an expect(...).toThrow here like the other test you've added? I think this could result in false positives if convertDispenseNotification doesn't throw an error

packages/coordinator/src/services/translation/request/dispense/dispense-notification.ts
```diff
@@ -67,6 +67,15 @@ export function convertDispenseNotification(bundle: fhir.Bundle, logger: pino.Lo
   ) as fhir.IdentifierExtension
 
   const BSAId = commissionedByExtension.valueIdentifier.value
+
+  if (!BSAId){
+    throw new processingErrors.InvalidValueError(
+      // eslint-disable-next-line max-len
```
MatthewPopat-NHS, line 73: These lines don't look that long, is this necessary?

packages/coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
natasafrgk: Addressed. Thank you!

packages/coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
```diff
@@ -243,6 +243,21 @@ describe("fhir MedicationDispense maps correct values in DispenseNotification wh
...
+describe("fhir MedicationDispense throws error for DispenseNotification", () => {
+  let dispenseNotification: fhir.Bundle
+  const testFileDir = "../../tests/resources/test-data/fhir/dispensing/"
+
+  test("missing value of ODS code for reimbursement authority", () => {
+    const testFileName = "Process-Request-Dispense-Notifications-No-Ods-Value-for-ReimbursementAuthority.json"
+    dispenseNotification = TestResources.getBundleFromTestFile(testFileDir + testFileName)
+    expect(() => {
+      convertDispenseNotification(dispenseNotification, logger)
+    }).toThrow(// eslint-disable-next-line max-len
```
MatthewPopat-NHS, line 255: again don't think this eslint comment is necessary

JackSpagnoliNHS, line 255: Following Matt's comment this comment could be removed too

packages/coordinator/src/services/translation/request/dispense/dispense-notification.ts
natasafrgk: You are correct. Removed.

packages/coordinator/tests/services/translation/request/dispense/dispense-notification.spec.ts
natasafrgk: You are correct. Removed.

JackSpagnoliNHS (approved):

PR #2563, opened 2024-07-22
Docs: [AEA-3930] - Updated mandatory fields for resources in spec

TomSmithCoding:
## Summary
- Routine Change

### Details
Added all listed mandatory requirements from ticket into the spec

### DIFF
--- a/packages/specification/schemas/components/MedicationRequest/CancelRequestMedicationRequest.yaml
+++ b/packages/specification/schemas/components/MedicationRequest/CancelRequestMedicationRequest.yaml
@@ -1,6 +1,7 @@
 type: object
 required:
   - statusReason
+  - extension
   - resourceType
   - identifier
   - status
@@ -122,6 +123,7 @@ properties:
     description: |
       A FHIR object containing information relevant to how the prescription should be dispensed. 
       These values should be consistent across all MedicationRequests in a Bundle.
+    required: [extension, quantity]
     properties:
       extension:
         type: array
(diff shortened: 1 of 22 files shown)


packages/specification/schemas/components/MedicationRequest/CancelRequestMedicationRequest.yaml
```diff
@@ -1,6 +1,7 @@
 type: object
 required:
   - statusReason
+  - extension
```
JackSpagnoliNHS, line 4: The extension field (line 34-38) currently contains only DMResponsiblePractitioner items, but example primary-care/acute/nominated-pharmacy/medical-prescriber/cancellations/1-Process-Request-Cancel-200_OK.json has an extension containing a Extension-DM-PrescriptionType, but no Extension-DM-ResponsiblePractitioner.

It might be worth double checking what that field can contain before we mark it as mandatory

packages/specification/schemas/components/_fragments/groupIdentifier.yaml
```diff
@@ -1,5 +1,5 @@
 type: object
-required: [value]
+required: [system, value, extension]
```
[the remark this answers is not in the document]
TomSmithCoding: Yep you are right, looks like I misread point 5 on the ticket to also include the extension as mandatory. Will remove this now.

packages/specification/schemas/components/MedicationRequest/CancelRequestMedicationRequest.yaml
TomSmithCoding: https://simplifier.net/packages/uk.nhsdigital.r4/2.4.5/files/566259/~details Looking down at responsiblePractitioner (eps-5) it looks like you might be right as it states to only supply this when the requester can't be used. So i'll revert this one too.

JackSpagnoliNHS (approved):

PR #2788, opened 2024-09-30
Fix: [AEA-0000] - enable sha256

anthony-nhs:
## Summary

- Routine Change

### Details

- enable sha256


MatthewPopat-NHS:
This was superceded by the per supplier toggle

PR #3041, opened 2024-12-30
Update: [AEA-4475] - Alter telecom wording

wildjames:
## Summary

- Routine Change

### Details

In our schema it currently says that in the telecom that phone, fax email and other can be applied and sent through to the EPS API. 

The change we want to apply is to "Phone" only.


seatonator (dismissed):

packages/specification/schemas/components/HealthcareService/HealthcareService.yaml
```diff
@@ -44,16 +44,16 @@ properties:
       type: object
       properties:
         system:
-          description: Means of communication, such as phone or email.
+          description: Mean of communication allowed is Phone only
```
MatthewPopat-NHS, line 47: It's still a means of communication even if it's singular.

packages/specification/schemas/components/HealthcareService/HealthcareService.yaml
```diff
@@ -44,16 +44,16 @@ properties:
...
+          enum: [phone]
           default: phone
         use:
           type: string
           description: Location associated with the contact details provided.
           enum: [home, work, temp, mobile]
           default: work
         value:
-          description: Phone number, email address, or other identifier for use with contact system.
+          description: Phone number for use with contact system.
```
MatthewPopat-NHS, line 57: "for use with contact system" doesn't really make sense if phone is the only system. Let just go with "The phone number"

MatthewPopat-NHS (changes requested):
Same comments apply across all 4 files

MatthewPopat-NHS (approved):

PR #3152, opened 2025-02-05
Fix: [AEA-4403] - use epsat to talk proxygen api

anthony-nhs:
## Summary

- Routine Change

### Details

- have config in epsat to talk to proxygen api
- pass an x-correlation-id from epsat client to epsat backend
- pass x-correlation-id from epsat backend to downstream api with a new x-request-id
- remove console.log
- add convert endpoint to PTL prescribing as its used by epsat
- set use mock signing on by default in dev


wildjames (changes requested):
Code all looks fine (plus, we had a guided tour as a team), happy to approve this once the test coverage has been expanded to meet our target.

wildjames (approved):

PR #3186, opened 2025-02-17
New: [AEA-0000] - use cdk

anthony-nhs:
## Summary

- Routine Change

### Details

- use cdk


cdk.json
```diff
@@ -0,0 +1,100 @@
...
+    "@aws-cdk/aws-kms:reduceCrossAccountRegionPolicyScope": true,
+    "@aws-cdk/aws-eks:nodegroupNameAttribute": true,
+    "@aws-cdk/aws-ec2:ebsDefaultGp3Volume": true,
+    "@aws-cdk/aws-ecs:removeDefaultDeploymentAlarm": true,
+    "@aws-cdk/custom-resources:logApiResponseDataPropertyTrueDefault": false,
+    "@aws-cdk/aws-s3:keepNotificationInImportedBucket": false,
+    "serviceName": "pd-dummy",
+    "VERSION_NUMBER": "foo",
+    "COMMIT_ID": "foo",
+    "accountId": "766544332211",
```
MatthewPopat-NHS, line 77: shouldn't this be "foo" too?

cdk.json
```diff
@@ -0,0 +1,100 @@
...
+    "logRetentionInDays": 30,
+    "vpcId": "vpc-53412ffeeddccbb12",
```
MatthewPopat-NHS, line 79: shouldn't this be "foo" too?

packages/cdk/resources/ECSTasks.ts
```diff
@@ -0,0 +1,173 @@
...
+        TARGET_SPINE_SERVER:  props.targetSpineServer,
+        MTLS_SPINE_CLIENT: "true",
+        PRESCRIBE_ENABLED: "true",
+        DISPENSE_ENABLED: "true",
+        COMMIT_ID: props.commitId,
+        CRL_DISTRIBUTION_DOMAIN: "crl.nhs.uk",
+        CRL_DISTRIBUTION_PROXY: "crl.nhs.uk",
+        DEPLOYED_VERSION: props.version,
+        DOSE_TO_TEXT_MODE: "AUDIT",
+        ENVIRONMENT: "internal-dev",
```
MatthewPopat-NHS, line 125: This should be passed in.

cdk.context.json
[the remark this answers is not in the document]
anthony-nhs: yes - its needed so that you can do a cdk synth without needing to have authorised against AWS. 
We run cdk synth in common quality checks job which does not have a connection to AWS

All the values in it are dummy values

cdk.json
anthony-nhs: left it as a 'number' so that it passes any checks for a valid account id. Its a dummy value and not one of our accounts

MatthewPopat-NHS (approved):

PR #3503, opened 2025-05-20
Docs: [AEA-3929] - Update schema examples to remove WhenPrepared and daysSupply objects

seatonator:
Removed WhenPrepared and DaysSupply

- Routine Change


seatonator:
Duplicate

PR #3903, opened 2025-09-24
Docs: [AEA-5748] - OAS - Medication request extension

kris-szlapa:
## Summary

🎫 [AEA-5748](https://nhsd-jira.digital.nhs.uk/browse/AEA-5748) OAS - Medication request extension

- Routine Change

### Details

This pull request updates the `ControlledDrug` extension to allow both the `quantityWords` and `schedule` elements to be present together, in line with the Simplifier specification.


MatthewPopat-NHS:
This is wrong but a common misunderstanding, oneOf means that each item in the array is one of these two type definition, not that the array should only contain oneOf these two things.

